PR #2, opened 2020-08-07
OpenConceptLab/ocl_issues#286 Deploy oclapi2 to QA

rkorytkowski:
Initial changes in preparation for QA deployment.


core/settings.py
```diff
@@ -104,15 +104,15 @@
...
+        'PASSWORD': os.environ.get('DB_PASSWORD', 'Postgres123'),
+        'HOST': os.environ.get('DB_HOST', 'db'),
         'PORT': os.environ.get('DB_PORT', 5432),
     }
 }
 
 ELASTICSEARCH_DSL = {
     'default': {
-        'hosts': os.environ.get('ES_URL', 'host.docker.internal:9200')
+        'hosts': [os.environ.get('ES_HOST', 'es:9200'), 'es1:9200']
```
rkorytkowski, line 115: In the next step I'll be adding an ngnix service to proxy and loadbalance requests instead of declaring nodes here.

PR #10, opened 2021-06-01
OpenConceptLab/ocl_issues#766 | API Versioning

snyaggarwal:
#### Tasks:
- [x] API_VERSION env var for major.minor.maintenance
- [x] API_BUILD env var for commit SHA (default `dev`)
- [x] `core.__version__` to reflect api version (with build)
- [x] All responses to have a custom header `X-OCL-API-VERSION` (using middleware)
- [x] `/` API to show version (with build) info with other info
- [x] `/version/` API to tell app version (with build)


core/settings.py
```diff
@@ -357,3 +359,7 @@
     EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 else:
     EMAIL_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
+
+API_VERSION = os.environ.get('API_VERSION', '2.0.1')
+API_BUILD = os.environ.get('API_BUILD', 'dev')
```
rkorytkowski, line 364: We don't want it to be a runtime variable rather a build time variable. In other words please hardcode version, build number and full version with build number in core/__init__.py and read it from core/settings.py. It needs to be set by a build script when running docker-compose build. I can do that part.

rkorytkowski (changes requested):

core/settings.py
snyaggarwal: Added version/build in core/__init__.py

rkorytkowski (approved):

PR #14, opened 2021-08-03
OpenConceptLab/ocl_issues#22 | exclude retired concepts and/or mappings from exports

PatrickCmd:
Retired concepts and/or mappings can now be excluded from concepts and or mappings.


core/common/tasks.py
```diff
@@ -56,7 +56,7 @@ def export_source(self, version_id):
     version.add_processing(self.request.id)
     try:
         logger.info('Found source version %s.  Beginning export...', version.version)
-        write_export_file(version, 'source', 'core.sources.serializers.SourceVersionExportSerializer', logger)
+        write_export_file(self.request, version, 'source', 'core.sources.serializers.SourceVersionExportSerializer', logger)
```
snyaggarwal, line 59: @PatrickCmd these methods should not be dependent on the `request` object.
passing `include_retired=True` in `export_source` directly as a parameter keeps things flexible and the same can be passed to `write_export_file`. The view should control request query params.
The reason I don't like using request object in `write_export_file` is it's a very low-level method and if I want to call it anywhere else I need to have a full request object, which is an overkill in my opinion.

PatrickCmd: @snyaggarwal I have worked on this. please do review again.

snyaggarwal:
@PatrickCmd The only thing missing is a test to exclude retired concepts

core/common/models.py
```diff
@@ -722,6 +722,11 @@ def is_exporting(self):
     def export_path(self):
         last_update = self.last_child_update.strftime('%Y%m%d%H%M%S')
         return self.generic_export_path(suffix="{}.zip".format(last_update))
+    
+    @cached_property
+    def exclude_retired_export_path(self):
+        last_update = self.last_child_update.strftime('%Y%m%d%H%M%S')
+        return self.generic_export_path(suffix="{}_non_retired.zip".format(last_update))
```
snyaggarwal, line 729: non_retired -> unretired

core/common/utils.py
```diff
@@ -330,6 +336,8 @@ def write_export_file(
     logger.info('Done compressing.  Uploading...')
 
     s3_key = version.export_path
+    if not include_retired:
```
snyaggarwal, line 339: `export_path` is an expensive call, since its calculates on runtime so use it only if needed
`s3_key = version.export_path if include_retired else version.exclude_retired_export_path`

snyaggarwal:
@PatrickCmd This looks good. Few things that are still left:
1. `GET /export/?includeRetired=true` should work as is (export_path) but `GET /export/?includeRetired=false` should use new `exclude_retired_export_path`
2. An integration test for `?includeRetired=false`, you can use the existing integration test `ExportSourceTaskTest.test_export_source`

core/common/models.py
PatrickCmd: updated

core/common/utils.py
PatrickCmd: addressed

core/collections/views.py
```diff
@@ -594,7 +594,8 @@ class CollectionVersionExportView(ConceptContainerExportMixin, CollectionVersion
     def handle_export_version(self):
         version = self.get_object()
         try:
-            export_collection.delay(version.id)
+            include_retired = parse_boolean_query_param(self.request, INCLUDE_RETIRED_PARAM)
+            export_collection.delay(version.id, include_retired)
```
snyaggarwal, line 598: @PatrickCmd we need to send True if the includeRetired param is not there or is not false

PatrickCmd: Addressed.

PR #15, opened 2021-08-05
OpenConceptLab/ocl_issues#878 | Locale type standardization

snyaggarwal:


core/concepts/migrations/0019_auto_20210805_0351.py
```diff
@@ -0,0 +1,39 @@
...
+    LocalizeText = apps.get_model('concepts', 'LocalizedText')
+
+    def batch_update(values_from, value_to):
+        batch_size = 1000
+        queryset = LocalizeText.objects.filter(type__in=values_from)
+        total = queryset.count()
+        for start in range(0, total, batch_size):
+            end = min(start + batch_size, total)
+            locales = LocalizeText.objects.filter(
+                id__in=queryset.order_by('id')[start:end].only('id').values_list('id', flat=True))
```
rkorytkowski, line 16: I feel this will not perform well... start:end is a slow one for millions of records to update as it uses OFFSET. Also if you use 'values_list' you don't need 'only'. We should be able to change it to an update without a subquery like:
update localized_texts set type='FULLY_SPECIFIED' where type in ('Fullly Specified', 'Fully_Specified', ...) don't we?

snyaggarwal: Ya, I can change that to run direct SQL:
`update localized_texts set type='FULLY_SPECIFIED' where type in ('Fullly Specified', 'Fully_Specified', ...)`

PR #20, opened 2021-08-27
OpenConceptLab/ocl_issues#897 Run OCL API using gunicorn

rkorytkowski:


requirements.txt
```diff
@@ -17,7 +18,7 @@ celery[redis]==4.4.7
 celery_once==3.0.1
 flower==0.9.5
 redis==3.5.3
-kombu==4.5.0
+kombu==4.6.11
```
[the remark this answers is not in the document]
snyaggarwal: I don't remember exactly the issues with upgrading it, but as far as I remember those issues were with latest kombu (version 5), upgrading minor should be ok

requirements.txt
```diff
@@ -26,4 +27,6 @@ coverage==5.3.1
 tblib==1.7.0
 django-ordered-model==3.4.1
 django-health-check>=3.16.2
-markdown==3.3.4
\ No newline at end of file
+markdown==3.3.4
+# Development
+git-pylint-commit-hook==2.5.1
```
snyaggarwal, line 32: @rkorytkowski do we want to add this in this pull request? I haven't tried this yet. Maybe we take it out from this pull request

rkorytkowski: Yeah, sorry my bad. I'll revert it. I was experimenting with the git hook and it sneaked through.

PR #43, opened 2021-09-10
OpenConceptLab/ocl_issues#965 | using whitenoise to serve staticfiles for non-dev mode (with gunicorn)

snyaggarwal:


core/urls.py
```diff
@@ -51,7 +50,6 @@
     url(r'^swagger/$', SchemaView.with_ui('swagger', cache_timeout=0), name='schema-swagger-ui'),
     url(r'^redoc/$', SchemaView.with_ui('redoc', cache_timeout=0), name='schema-redoc'),
     path('healthcheck/', include('core.common.healthcheck.urls')),
-    path('admin/', admin.site.urls, name='admin_urls'),
```
rkorytkowski, line 54: Why are you removing this?

core/settings.py
```diff
@@ -266,6 +265,9 @@
 # https://docs.djangoproject.com/en/3.0/howto/static-files/
 
 STATIC_URL = '/static/'
+PROJECT_DIR = os.path.dirname(os.path.abspath(__file__))
+STATIC_ROOT = os.path.join(PROJECT_DIR, 'static')
```
rkorytkowski, line 269: No need to call collectstatic in Dockerfile?

core/urls.py
snyaggarwal: we don't use any admin routes/features. And also, with whitenoise all the calls were getting redirected to Django Login.

core/settings.py
snyaggarwal: ya collectstatic wasn't needed.

rkorytkowski:
Thanks @snyaggarwal. Committed applying a few of your ideas at https://github.com/OpenConceptLab/oclapi2/commit/2f6df521d077e7887b094e1da403eb76d56ced8a

snyaggarwal:
Closing this pull request as this has been added on master already.

PR #240, opened 2022-05-10
Valueset refactoring

snyaggarwal:
1. Using CollectionReferenceParser for the compose statements
2. Formatting tests/serializers

### DIFF
--- a/core/value_sets/serializers.py
+++ b/core/value_sets/serializers.py
@@ -2,11 +2,11 @@ from rest_framework import serializers
 from rest_framework.fields import CharField, DateField, SerializerMethodField, ChoiceField
 
 from core.code_systems.serializers import CodeSystemConceptSerializer
-from core.collections.models import Collection, CollectionReference
+from core.collections.models import Collection
+from core.collections.parsers import CollectionReferenceParser
 from core.collections.serializers import CollectionCreateOrUpdateSerializer
 from core.common.serializers import StatusField, IdentifierSerializer, ReadSerializerMixin
 from core.orgs.models import Organization
-from core.sources.models import Source
 from core.users.models import UserProfile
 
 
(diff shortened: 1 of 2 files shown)


core/value_sets/serializers.py
```diff
@@ -258,4 +200,4 @@ def get_resource_type(_):
 
     @staticmethod
     def get_meta(obj):
-        return {'lastUpdated': obj.updated_at}
+        return dict(lastUpdated=obj.updated_at)
```
rkorytkowski, line 203: Is this preferred style?

core/value_sets/tests.py
```diff
@@ -138,52 +165,55 @@ def test_create_with_filter_and_system(self):
...
         resource = response.data
         self.assertEqual(resource['version'], '0.1')
-        self.assertEqual(resource['identifier'][0]['value'], '/users/' + self.user.mnemonic
-                         + '/ValueSet/c2/')
+        self.assertEqual(resource['identifier'][0]['value'], f'/users/{self.user.username}/ValueSet/c2/')
         self.assertEqual(len(resource['compose']['include']), 1)
-        self.assertEqual(resource['compose']['include'][0]['system'], '/some/url')
+        self.assertEqual(resource['compose']['include'][0]['system'], 'http://some/url')
         self.assertEqual(resource['compose']['include'][0]['version'], self.org_source_v2.version)
         self.assertEqual(len(resource['compose']['include'][0]['concept']), 1)
```
rkorytkowski, line 179: It would be great to have some compare util for response.data which takes an expected json, disregards ordering and can be set to ignore certain fields.

rkorytkowski: Verifying response would be much cleaner and more accurate.

rkorytkowski (approved):

core/value_sets/serializers.py
snyaggarwal: Don't think there is any recommendation between the two. This is just my personal preference, it feels more readable.

PR #278, opened 2022-10-18
OpenConceptLab/ocl_issues#1387 | Celery Beat

snyaggarwal:


rkorytkowski:
@snyaggarwal I did further reading on celery beat and in default configuration it stores the schedule and history of past runs in a local file (within a container). It's not suitable for deployment in a cluster as we will loose history of runs with every restart. We could add a persistent volume, but then we will need to carefully handle it and deploy always to the same AWS zone and we would loose the volume in case of zone failure. I think we need to use django-celery-beat extension so that schedule and most importantly history of past runs is stored in a DB. This way we can freely deploy celery beat to container in any AWS zone and have it read schedule from DB. Do you see any drawbacks? Have you used django-celery-beat in the past? It seems to be well maintained and still active from GH history https://github.com/celery/django-celery-beat

start_celery_beat.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+./wait_for_it.sh ${REDIS_HOST}:${REDIS_PORT} -t 0
+
+celery beat -A core.celeryconf -l info
```
rkorytkowski, line 7: Shouldn't this be "... -A core.celery"?

snyaggarwal: Fixed

snyaggarwal:
@rkorytkowski This make sense. I have never used django-celery-beat. I will give it a try and write a sudo job to test it out. Will update PR with it.

rkorytkowski:
As far as healthcheck is concerned it would be best to have a dumb task scheduled to be run every 10 seconds or so and and oclapi healthcheck endpoint returning an error if the task wasn't run for more than e.g. 15 seconds.

rkorytkowski:
I like this approach for container healthcheck: https://github.com/celery/celery/issues/3694#issuecomment-481114772

Basically aside from the api endpoint healthcheck I suggested above, add a django management command to check the same so it can be called directly from the container.

snyaggarwal:
@rkorytkowski this requires celery upgrade (celery5). I will give that a try first in the same PR

snyaggarwal:
@rkorytkowski we are stuck in the upgrade loop:
1. django-celery-beat needs celery5 (we are on celery4) -- django-celery-beat's older version that supports celery4 doesn't work with django4 (we are on django4)
2. celery5 works fine (with a couple of small changes in the command line).
3. We are currently on flower 0.9.5 which doesn't work with celery5, so we need to upgrade to flower 1.0.1+
3. flower 1.0.1+ has this [open issue (an PR)](https://github.com/mher/flower/issues/1231) that stops the API to be used if auth is enabled :(

So in-short we (read flower) are not ready for django-celery-beat yet. I am hoping the PR on flower will be merged soon.

snyaggarwal:
@rkorytkowski the temp option could be to fork Flower, merge the fix (PR), and use it from there until its merged and released on the upstream.

snyaggarwal:
@rkorytkowski I have added healthcheck described in https://github.com/celery/celery/issues/3694#issuecomment-481114772
- A scheduled job `core.common.tasks.beat_healthcheck` that runs every minute and sets a key `settings.CELERYBEAT_HEALTHCHECK_KEY` in Redis (with value datetime) and expires in 120 seconds.
- A management command `celery_beat_healthcheck` that checks the existence of that key and exits (1) if the key doesn't exist.

The remaining bit is to initiate the management command (`python manage.py celery_beat_healthcheck`) periodically (every minute, starting two minutes from when the celery-beat starts running), can you take care of that?

rkorytkowski:
> The remaining bit is to initiate the management command (`python manage.py celery_beat_healthcheck`) periodically (every minute, starting two minutes from when the celery-beat starts running), can you take care of that?

Thanks! Yes, I'll do that!

snyaggarwal:
this was merged manually. Closing this.

PR #318, opened 2022-12-06
OpenConceptLab/ocl_issues#1399 | names/descriptions migrations to split names and descriptions

snyaggarwal:


core/concepts/migrations/0039_auto_20221122_0552.py
```diff
@@ -0,0 +1,49 @@
...
+                    external_id=localized_text.external_id,
+                    name=localized_text.name,
+                    type=localized_text.type,
+                    locale=localized_text.locale,
+                    locale_preferred=localized_text.locale_preferred,
+                    created_at=localized_text.created_at,
+                ))
+
+    LocalizedText.objects.bulk_create(names, batch_size=1000)
+    Concept.names.through.objects.filter(id__in=concepts_names_to_delete).delete()
```
rkorytkowski, line 38: I would be surprised if __in was not limited to a few thousand items in an array... A quick fix would be to do it in batches as well.

core/concepts/migrations/0039_auto_20221122_0552.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('concepts', '0038_auto_20221122_0522'),
+    ]
+
+    operations = [
+        migrations.RunPython(add_names_for_latest_version)
```
rkorytkowski, line 48: Is a migration run in a single transaction? If so, then it may be too much to run in a single transaction.

core/concepts/models.py
```diff
@@ -141,6 +86,75 @@ def is_search_index_term(self):
...
+        )
+
+    @classmethod
+    def get_dormant_queryset(cls):
+        return cls.objects.filter(name_locales__isnull=True, description_locales__isnull=True)
+
+    @classmethod
+    def dormants(cls, raw=True):
+        if raw:
+            with connection.cursor() as cursor:
```
rkorytkowski, line 142: We are not able to use a cursor outside of celery indexing task. Do we need another celery worker connecting directly to postgres? When do we want to run it?

rkorytkowski (changes requested):
Have you done any initial performance testing of the migration?

Could you please list migration steps in a few bullet points in a comment on this PR? It would be easier for me to go through the code and verify if it does what you intended and if there's some optimisation that can be done?

core/concepts/models.py
snyaggarwal: We wont need this piece of code after we migrate. This was an api for admin user to get list of dormant locales. Which never really worked on Staging.

rkorytkowski:
@snyaggarwal How big was the localized_text table and a new description table?

snyaggarwal:
@rkorytkowski Around 300k localized_texts

snyaggarwal:
@rkorytkowski I took QA db dump and tested these migrations (without any code changes), the numbers are below:
localized_texts -> 599365
concepts_names -> 536753 (many-to-many)
concepts_descriptions -> 70914 (many-to-many)

  Applying concepts.0038_auto_20221122_0522...Time:  28.620164394378662
 OK
  Applying concepts.0039_auto_20221122_0552...Time:  153.2996428012848
 OK
  Applying concepts.0040_auto_20221122_0613...Time:  123.85713315010071
 OK

I am not sure what happened with Dev environment, I think something while multiple deployments failed on Dev.
But these numbers are consistent from just CIEL dictionary migration on my local and QA data migration on local.

rkorytkowski:
@snyaggarwal yeah, DB might have been executing queries even though services were restarted. Anyway if we just assume that it is linear time increase we will wait for more than 2 weeks for 3 billion records. We still need to go down with this number.

snyaggarwal:
Agreed

snyaggarwal:
@rkorytkowski with raw queries on QA data -- 

  Applying concepts.0038_auto_20221122_0522...Time:  17.52713656425476
 OK
  Applying concepts.0039_auto_20221122_0552...Time:  56.87551474571228
 OK
  Applying concepts.0040_auto_20221122_0613...Time:  7.51812481880188
 OK

snyaggarwal:
@rkorytkowski I took db dump from Staging (without mappings/collections tables).
Below were the numbers (3billion number that we saw was the size of localized_texts table in Bytes  ):
```
postgres=# select count(*) from localized_texts;
  count
---------
 8881141
(1 row)

postgres=# select count(*) from concepts_names;
  count
----------
 10542441
(1 row)

postgres=# select count(*) from concepts_descriptions;
 count
--------
 286273
(1 row)
postgres=# select count(*) from concepts;
  count
---------
 4039783
(1 row)

```

Ran the migrations, below were the timings:

rkorytkowski:
What an epic mistake! I must have ran pg_relation_size and not count. That's a relief! 32 minutes is definitely acceptable. We either need to take the service down for that period to prevent modifications or run the migration for the second time after new code is deployed to migrate any modifications done in that half an hour. The second run should be much faster. What do you think?

snyaggarwal:
@rkorytkowski I would suggest that we do this with downtime (no new writes), because of two reasons, first migration will only run once and second one of the migrations run on everything, doesn't exclude already processed.

rkorytkowski:
All right! Let's plan for that in January.

rkorytkowski (approved):
Thanks Sunny! All good! Go ahead and run it on QA.

snyaggarwal:
Squashed and merged manually.

PR #510, opened 2023-06-19
OpenConceptLab/ocl_issues#1597 Migrate from ES to OpenSearch

rkorytkowski:
@snyaggarwal please try it out locally. You need to drop volumes (docker-compose down -v) and create everything from scratch.

The migration path will be as follows:

0. Setup OpenSearch for all envs on AWS with terraform
1. Push an image with OpenSearch libs to dockerhub built from this branch (do not replace master images yet)
2. Deploy migration service using the opensearch image on QA
3. Once the migration is done. Merge opensearch branch with master and deploy to QA.
4. Test searching. 
5. Deploy migration service using the opensearch image on staging
6. Deploy api to staging
7. Test searching.
8. Repeat 5-7 for production.
9. Bring down ES services
10. Rebuild indexes again on staging and production just in case any updates happened between migration and switching API. @snyaggarwal, is the rebuild operation completely breaking search results or is it handled in a non-disruptive way? 

The process implies that we will not be able to deploy anything from the master branch until done with migration. I was hoping to do all that next week.


rkorytkowski:
The switch from elasticsearch libs to opensearch libs is required since elasticsearch libs are purposefully and artificially (for most cases) made incompatible with opensearch. Not sure if the other way is the same case.

snyaggarwal:
@rkorytkowski I couldn't find any documentation for async indexing. The documentation for [signals](https://django-opensearch-dsl.readthedocs.io/en/latest/signals/) also just have post_index.

snyaggarwal:
@rkorytkowski Also I checked there is only a command to rebuild the indexes (along with create and delete) but no command to just populate.
https://django-opensearch-dsl.readthedocs.io/en/latest/management/

snyaggarwal:
> @rkorytkowski Also I checked there is only a command to rebuild the indexes (along with create and delete) but no command to just populate. https://django-opensearch-dsl.readthedocs.io/en/latest/management/

Well ignore, there is one.

snyaggarwal:
@rkorytkowski Search is failing right now with error `TypeError: unhashable type: 'Bool'`

snyaggarwal:
> @rkorytkowski Search is failing right now with error `TypeError: unhashable type: 'Bool'`

@rkorytkowski I tried to read the documentation of OpenSearch, and looks like we will have to change the code to use OpenSearch-dsl lang

PR #538, opened 2023-07-19
Add db user and name to env variable

chojisan:
Added DB_DATABASE and DB_USER on environment variable in case user wants to use a separate postgres database deployment


core/settings.py
```diff
@@ -184,8 +184,8 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql',
-        'NAME': os.environ.get('DB', 'postgres'),
```
snyaggarwal, line 187: Keep this var as `DB`

PR #545, opened 2023-07-25
chore: make scheme configurable for elastic search connection

Salaton:
This PR introduces enhancements to the `es_get` function, allowing the scheme used for connecting to the Elasticsearch cluster to be configurable. This is done to address the error `Client sent an HTTP request to an HTTPS server` when passing `http` in the host while using Elasticsearch on elastic cloud (https://www.elastic.co/cloud/)


rkorytkowski:
Thank you for your contribution!

PR #581, opened 2023-08-31
Switch response body to OperationOutcome for empty queryset in $lookup

IanMinash:
Noticed that when an empty queryset is returned in a `$lookup` operation the response body is DRF's default `404` response instead of an `OperationOutcome` as documented in the [$lookup docs](https://www.hl7.org/fhir/codesystem-operation-lookup.html)


rkorytkowski:
Thank you for the contribution!

PR #643, opened 2024-01-24
OpenConceptLab/ocl_issues#1729 | Azure Blob Storage Class for exports and other uploads

snyaggarwal:
- Addresses OpenConceptLab/ocl_issues#1729
- Implements `CloudStorageServiceInterface` in `cloud.azure.BlobStorage`

### DIFF
--- a/core/services/storages/cloud/azure.py
+++ b/core/services/storages/cloud/azure.py
@@ -66,12 +66,13 @@ class BlobStorage(CloudStorageServiceInterface):
 
         return None
 
-    def upload_file(
-            self, key, file_path=None, headers=None, binary=False, metadata=None
-    ):  # pylint: disable=too-many-arguments
-        read_directive = 'rb' if binary else 'r'
-        file_path = file_path if file_path else key
-        return self._upload(key, open(file_path, read_directive).read(), headers or metadata)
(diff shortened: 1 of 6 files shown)


core/services/storages/cloud/azure.py
```diff
@@ -0,0 +1,133 @@
...
+            pass
+
+        return None
+
+    def upload_file(
+            self, key, file_path=None, headers=None, binary=False, metadata=None
+    ):  # pylint: disable=too-many-arguments
+        read_directive = 'rb' if binary else 'r'
+        file_path = file_path if file_path else key
+        return self._upload(key, open(file_path, read_directive).read(), headers or metadata)
```
rkorytkowski, line 74: One thing is that it is very easy to make it using a stream and not load the whole file into memory, because the code will fail as soon as we start creating exports that do not fit into memory. I see that upload_blob supports passing a file handler open(file_path, read_directive) without calling read(). See https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-upload-python#upload-a-block-blob-from-a-local-file-path

rkorytkowski:
Thanks @snyaggarwal !

PR #838, opened 2026-04-06
Create GraphQL service permission archtecture

### LINKED ISSUE #2356: Defer GraphQL permission decorators until group authorization service is complete
## User story
As a developer, I want to postpone implementing GraphQL permission handling (needed) for GraphQL queries so I can avoid merge conflicts while the group authorization/permissions service is still under active development.

## Requirements
* Do not start GraphQL permission work for GraphQL until the group authorization/permissions work for OCL premium services is completed.
* Once unblocked, implement GraphQL permissions as decorators in Strawberry (GraphQL) integrated with Django (no Django REST framework classes available).
* Align the approach and sequencing with prior discussions with @snyaggarwal @paynejd .

## Acceptance criteria
- [ ] Work on GraphQL permission decorators for GraphQL is explicitly blocked until the group authorization service is merged/completed.
- [ ] When unblocked, permissions are implemented using Strawberry decorators (not DRF permission classes).

filiperochalopes:
## Linked Issue

Closes https://github.com/OpenConceptLab/ocl_issues/issues/2356


core/concepts/search.py
```diff
@@ -4,8 +4,192 @@
...
+
+def get_concept_mandatory_exclude_words_criteria(query, include_map_codes=True):
+    """Build the excluded-word wildcard clauses shared by REST and GraphQL."""
+    criterion = None
+    for must_not_have in CustomESSearch.get_must_not_haves(normalize_concept_search_query(query)):
+        criteria, _ = get_concept_wildcard_search_criterion(
+            f"{must_not_have}*",
+            include_map_codes=include_map_codes,
+        )
+        criterion = criteria if criterion is None else criterion | criteria
```
snyaggarwal, line 106: `get_concept_mandatory_exclude_words_criteria` uses `|` where the old code used `&`


`criterion = criteria if criterion is None else criterion | criteria  # new (|)`
vs original in views.py:
`criterion = criteria if criterion is None else criterion & criteria  # old (&)`
The downstream callers do `search.filter(~must_not_criterion)`, so:
`|` → excludes docs matching any excluded word (stricter, likely correct semantically)
`&` → excludes docs matching all excluded words simultaneously (the old REST behavior)
This silently changes REST concept search too, since `views.py:get_mandatory_exclude_words_criteria` now delegates here for concept documents.

core/graphql/queries.py
```diff
@@ -507,6 +570,10 @@ async def concepts(  # pylint: disable=too-many-arguments,too-many-locals
                 source_version,
                 pagination,
                 mapping_prefetch,
+                owner=owner_value,
+                owner_type=owner_type,
+                version_label=version or HEAD if source_version else None,
```
snyaggarwal: to make it less ambiguous
`(version or HEAD) if source_version else None`

core/graphql/queries.py
```diff
@@ -353,27 +380,34 @@ def concept_ids_from_es(
...
     if not trimmed:
         return [], 0
 
     try:
         search = ConceptDocument.search()
+        search = search.filter('term', retired=False)
         if source_version:
-            search = search.filter('term', source=source_version.mnemonic.lower())
-            if source_version.is_head:
+            search = search.filter('term', source=source_version.mnemonic)
```
snyaggarwal: `.lower()` is required since in ES
`source = fields.KeywordField(attr='parent_resource', normalizer="lowercase")`

core/graphql/permissions.py
```diff
@@ -0,0 +1,136 @@
...
+    owner_name = resolver.__qualname__.split('.', 1)[0]
+    owner_class = resolver.__globals__.get(owner_name)
+    if owner_class is None:
+        raise GraphQLError('Resolver permission target is not available.')
+    return owner_class()
+
+
+async def ensure_can_view_repo(user, source_version) -> None:
+    """Raise a GraphQL forbidden error when the repository is not visible to the user."""
+    request = SimpleNamespace(user=user)
```
snyaggarwal: This fakes a DRF request object to reuse a REST permission class in a GraphQL context. 
It works today but it's fragile — if CanViewConceptDictionary ever accesses anything else on the request, it silently breaks. The cleaner alternative is a thin shared function that takes (user, source_version) directly.

core/graphql/views.py
```diff
@@ -104,3 +107,13 @@ def make_invalid(auth_status='invalid'):
         context.auth_status = 'valid' if getattr(user, 'is_authenticated', False) else 'invalid'
 
         return context
+
+    async def execute_operation(self, request, context, root_value, sub_response):
```
snyaggarwal: from Claude:


Overriding execute_operation on the view to return an ExecutionResult directly is not a documented Strawberry extension point — it's reaching into internals. The documented hook for this is process_errors on the schema (which the PR already uses in OCLGraphQLSchema). The auth short-circuit could instead be done via a middleware or the schema's execute hook, or simply by relying on the decorator check (which already runs before any resolver logic).

core/concepts/search.py
filiperochalopes: Confirmed and kept intentionally: `criterion | criteria` makes excluded terms remove documents matching any excluded word, which is the expected shared REST/GraphQL behavior after extracting this helper.

core/graphql/queries.py
filiperochalopes: Updated for clarity: `version_label=(version or HEAD) if source_version else None` now makes the precedence explicit.

core/graphql/queries.py
filiperochalopes: Fixed in the ES filters: `source=source_version.mnemonic.lower()` now matches the lowercase keyword normalizer. I also normalized the owner filter with `owner=owner.lower()`.

core/graphql/permissions.py
filiperochalopes: Refactored as suggested: `user_can_view_concept_dictionary(user, source_version)` now replaces the fake DRF request and keeps REST/GraphQL on the same shared rule.

core/graphql/views.py
filiperochalopes: Moved this out of the view override: `AuthStatusExtension.on_execute` now short-circuits invalid auth, and the custom `execute_operation` override was removed.

core/common/permissions.py
```diff
@@ -42,16 +42,29 @@ def has_object_permission(self, request, view, obj):
...
     """
     The user can view this source
     """
 
     def has_object_permission(self, request, view, obj):
-        if obj.public_access in [ACCESS_TYPE_EDIT, ACCESS_TYPE_VIEW]:
-            return True
-
-        return super().has_object_permission(request, view, obj)
+        return user_can_view_concept_dictionary(request.user, obj)
```
snyaggarwal, line 67: this now never calls super and `HasPrivateAccess` is dead code

filiperochalopes: Good catch on the inheritance — CanViewConceptDictionary no longer calls super(), so the HasPrivateAccess base is meaningless here. Fixed by changing it to inherit from BasePermission directly.

Note: `HasPrivateAccess` itself is still in use elsewhere (core/orgs/views.py:116,127,137, core/common/mixins.py:407, and `CanEditConceptDictionary` still extends and calls super()), so the class is not dead globally — only the inheritance link from CanViewConceptDictionary was.

PR #845, opened 2026-04-09
OpenConceptLab/ocl_online#51 | Block Anonymous API usage except from approved clients

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_online#51

### DIFF
--- a/core/middlewares/middlewares.py
+++ b/core/middlewares/middlewares.py
@@ -108,7 +108,6 @@ class RequireAuthenticationMiddleware(BaseMiddleware):
         '/admin/',
     )
     exempt_exact_paths = {
-        '',
         '/',
         '/version',
         '/changelog',
@@ -135,13 +134,9 @@ class RequireAuthenticationMiddleware(BaseMiddleware):
             return True
 
         user = getattr(request, 'user', None)
-        return any((
-            getattr(user, 'is_authenticated', False),
-            self.is_exempt_path(request.path),
-            self.has_approved_client_header(request),
-            self.has_approved_api_key(request),
-            self.has_approved_ip(request),
-        ))
+        return getattr(user, 'is_authenticated', False) or self.is_exempt_path(request.path) or \
+            self.has_approved_client_header(request) or \
+            self.has_approved_api_key(request) or self.has_approved_ip(request)
 
     @classmethod
     def is_exempt_path(cls, path):
@@ -177,8 +172,7 @@ class RequireAuthenticationMiddleware(BaseMiddleware):
 
         authorization = request.META.get('HTTP_AUTHORIZATION', '').strip()
         x_api_key = request.META.get('HTTP_X_API_KEY', '').strip()
-        query_api_key = request.GET.get('api_key', '').strip()
-        tokens = [authorization, x_api_key, query_api_key]
+        tokens = [authorization, x_api_key]
         bearer_token = authorization.split(None, 1)[1].strip() if ' ' in authorization else ''
         if bearer_token:
             tokens.append(bearer_token)
@@ -187,18 +181,16 @@ class RequireAuthenticationMiddleware(BaseMiddleware):
 
     @staticmethod
     def has_approved_ip(request):
-        """Match source IPs using forwarded addresses first, then the socket address."""
+        """Match source IPs using the socket address only."""
         approved_ips = settings.APPROVED_ANONYMOUS_IPS
         if not approved_ips:
             return False
 
-        forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR', '')
-        remote_addr = request.META.get('REMOTE_ADDR', '')
-        ip_candidates = [ip.strip() for ip in forwarded_for.split(',') if ip.strip()]
-        if remote_addr:
-            ip_candidates.append(remote_addr.strip())
+        remote_addr = request.META.get('REMOTE_ADDR', '').strip()
+        if not remote_addr:
+            return False
 
-        return any(ip in approved_ips for ip in ip_candidates)
+        return remote_addr in approved_ips
 
 
 class FhirMiddleware(BaseMiddleware):
--- a/core/middlewares/middlewares.py
+++ b/core/middlewares/middlewares.py
@@ -2,7 +2,9 @@ import logging
 import time
 
 import requests
+from django.conf import settings
 from django.http import HttpResponseNotFound, HttpResponse
+from django.http.response import JsonResponse
 from django.utils.deprecation import MiddlewareMixin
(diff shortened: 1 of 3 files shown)


core/middlewares/middlewares.py
```diff
@@ -89,6 +90,117 @@ def __call__(self, request):
...
+        return any(token and token in approved_keys for token in tokens)
+
+    @staticmethod
+    def has_approved_ip(request):
+        """Match source IPs using forwarded addresses first, then the socket address."""
+        approved_ips = settings.APPROVED_ANONYMOUS_IPS
+        if not approved_ips:
+            return False
+
+        forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR', '')
```
paynejd: **Security: `X-Forwarded-For` is trivially spoofable.** Any anonymous client can send `X-Forwarded-For: <approved-ip>` and bypass auth entirely. This is only safe if a trusted reverse proxy (ALB, nginx) strips/overwrites the header before it reaches Django.

Options:
1. Document this assumption clearly (that the deployment must strip `X-Forwarded-For`)
2. Only trust `REMOTE_ADDR` (the socket address)
3. Use a proxy-aware IP resolution approach (e.g. Django's `SECURE_PROXY_SSL_HEADER` pattern or `django-ipware`)

This is the most important issue in this PR — it's a real auth bypass if the deployment doesn't strip the header upstream.

core/middlewares/middlewares.py
```diff
@@ -89,6 +90,117 @@ def __call__(self, request):
...
+    @staticmethod
+    def has_approved_api_key(request):
+        """Match allowlisted anonymous API keys from common request locations."""
+        approved_keys = settings.APPROVED_ANONYMOUS_API_KEYS
+        if not approved_keys:
+            return False
+
+        authorization = request.META.get('HTTP_AUTHORIZATION', '').strip()
+        x_api_key = request.META.get('HTTP_X_API_KEY', '').strip()
+        query_api_key = request.GET.get('api_key', '').strip()
```
paynejd: **Security: API keys in query strings end up in server access logs, browser history, referrer headers, and CDN logs.** Consider whether `?api_key=` is actually needed as a bypass vector, or if header-only (`X-API-KEY` / `Authorization`) is sufficient. If it must stay, document the risk.

core/middlewares/middlewares.py
```diff
@@ -89,6 +90,117 @@ def __call__(self, request):
...
+
+        return JsonResponse(self.forbidden_response, status=403)
+
+    def is_request_allowed(self, request):
+        """Return whether the current request can bypass authentication enforcement."""
+        if request.method == 'OPTIONS' or request.META.get('HTTP_USER_AGENT', '').startswith('ELB-HealthChecker'):
+            return True
+
+        user = getattr(request, 'user', None)
+        return any((
```
paynejd: **Nit: eager evaluation.** `any()` on a tuple evaluates all checks before short-circuiting — so `has_approved_ip` runs even if the user is authenticated. Use `or` chains for true short-circuit:

```python
return (
    getattr(user, 'is_authenticated', False)
    or self.is_exempt_path(request.path)
    or self.has_approved_client_header(request)
    or self.has_approved_api_key(request)
    or self.has_approved_ip(request)
)
```

core/middlewares/middlewares.py
```diff
@@ -89,6 +90,117 @@ def __call__(self, request):
...
+        '/users/password/reset/',
+        '/users/oidc/',
+        '/oidc/',
+        '/fhir/',
+        '/swagger',
+        '/redoc/',
+        '/admin/',
+    )
+    exempt_exact_paths = {
+        '',
```
paynejd: **Dead code:** `''` in `exempt_exact_paths` is unreachable. The normalization on line 149 does `path.rstrip('/') or '/'`, so an empty string is always coerced to `'/'`. You can remove the `''` entry.

core/settings.py
```diff
@@ -49,6 +49,17 @@
 else:
     ENABLE_THROTTLING = os.environ.get('ENABLE_THROTTLING', False) in ['true', 'True', 'TRUE', True]
 
+
+def get_list_from_env(name):
+    """Return a trimmed list for comma-separated environment variables."""
+    return [value.strip() for value in os.environ.get(name, '').split(',') if value.strip()]
```
paynejd: **Suggestion:** Return a `set` instead of a `list` — these are used for membership checks (`in`), and a set is both semantically correct and O(1):

```python
def get_list_from_env(name):
    return {value.strip() for value in os.environ.get(name, '').split(',') if value.strip()}
```

(You may want to rename to `get_set_from_env` accordingly.)

core/middlewares/tests.py
```diff
@@ -0,0 +1,137 @@
...
+        paths = [
+            '/users/alice/verify/token-123/',
+            '/users/alice/sso-migrate/',
+            '/users/alice/following/',
+        ]
+
+        for path in paths:
+            with self.subTest(path=path):
+                response = self.middleware(self.make_request(path))
+                self.assertEqual(response.status_code, 200)
```
paynejd, line 158: **Missing test coverage:**

1. **No test for `?api_key=` query string bypass** — header-based API key is tested but the query param path isn't.
2. **No test for `X-Forwarded-For` IP bypass** — only `REMOTE_ADDR` is tested. Given the spoofing concern, this deserves a test (and documents the expected behavior).
3. **No test for empty/whitespace header values** — e.g. `X-OCL-CLIENT: '  '` should be rejected after `.strip()`.
4. **No test for `REQUIRE_AUTHENTICATION=False`** — confirming the middleware isn't in the chain when disabled.

paynejd (changes requested):
Overall the design is clean and well-structured — each bypass mechanism is a separate method, easy to test and extend. Good test coverage too. A few things to address before merging, most importantly the X-Forwarded-For spoofing risk.

**Typo in PR title:** "Anonymouse" → "Anonymous"

paynejd (approved):
All six review comments addressed cleanly:

- **X-Forwarded-For spoofing** — resolved by only trusting `REMOTE_ADDR`
- **Query string API keys** — removed; header-only (`X-API-KEY` / `Authorization`)
- **Short-circuit evaluation** — `or` chains instead of `any()` tuple
- **Dead `''` in exempt_exact_paths** — removed
- **`get_set_from_env`** — renamed + returns a set for O(1) lookups
- **Test coverage gaps** — all four added (query param rejection, forwarded IP rejection, whitespace header, disabled setting)

Minor nit: first commit message still has "Anonymouse" typo — not blocking.

LGTM 👍

PR #847, opened 2026-04-13
code-review-graph (code-review-graph.com) in AGENTS.md

snyaggarwal:
https://code-review-graph.com/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -309,3 +309,5 @@ local*.sh
 local*.py
 setup/
 /.claude
+# Added by code-review-graph
+.code-review-graph/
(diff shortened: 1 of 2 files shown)


filiperochalopes (changes requested):
I think we should add the link for desired dependencies for development somewhere, maybe CONTRIBUTION.md just adding https://code-review-graph.com/#install

paynejd:
My opinion: this looks like a useful tool, but doesn't belong in our open-source distro. We need to create a convention around this, but `AGENTS.md` in each of our open source repo distros should be only what is required/necessary/unique -- we shouldn't be specifying implementer and dev tool stacks in this way.

My recommendation is to keep this in your own local AGENTS.md, specifically in your global AGENTS.md (that applies across all claude code and codex sessions) and simply specify there which repos it is relevant too. After evaluating it, if the team decides we want to all adopt it, it still doesn't belong in oclapi2/AGENTS.md, it belongs in a shared artifact repo. There are a bunch of git-friendly methods for managing AGENTS.md, CLAUDE.md, etc. for particular workspaces and globally that still take advantage of versioning.

paynejd (changes requested):
Only do the .gitignore. The AGENTS.md content belongs in local dev AGENTS.md (likely in your global AGENTS.md so that it applies across claude code/codex sessions). See my comment on the main thread.

filiperochalopes:
It makes sense @paynejd

PR #849, opened 2026-04-13
Improve development environment defaults and storage setup

filiperochalopes:
## Summary

This PR improves the local development environment by documenting the expected defaults, reducing maintenance overhead from deprecated components, and making the storage layer more practical for development and testing workflows.

## Context

The current development setup had a few limitations:

- MinIO is now obsolete in this project context and will no longer be maintained
- the development override file should not be treated as a documented default entrypoint, since it is an override-only file meant for local use
- using an AWS S3 bucket in development prevents some features from being tested properly or consistently
- a more efficient local storage service was needed to replace the previous approach

To address this, the development environment now moves toward a documented default `.env` template and a backup-based override workflow, while introducing `RUSTFS` as the preferred local S3-compatible service.

## What changed

- added `.env.example` documenting the expected development defaults for:
  - core services
  - storage configuration
  - optional integrations

- renamed `docker-compose.override.yml` to `docker-compose.override.yml.bak`

- updated the repository to ignore the active override file, so local developer-specific overrides are not committed accidentally

- preserved the backup override template so developers can still enable it locally when needed with:

~~~sh
cp docker-compose.override.yml.bak docker-compose.override.yml
~~~

- kept the expanded storage-related development setup available in the backup override file

- introduced `RUSTFS` as the local storage service, replacing the previous dependence on MinIO for active development workflows

## Why this approach

This approach keeps the repository cleaner and makes the development workflow more explicit:

- documented defaults live in `.env.example`
- local override behavior remains available, but is treated as local-only customization
- deprecated infrastructure is no longer presented as the default path
- local storage testing becomes more realistic and less constrained than using AWS S3 directly during development

## Benefits

- clearer onboarding for development environments
- less confusion around which files are documented defaults versus local overrides
- fewer accidental commits of machine-specific override changes
- better support for testing storage-dependent features locally
- improved maintainability by moving away from MinIO
- more efficient local S3-compatible behavior through `RUSTFS`

## Notes for reviewers

Developers who need local override behavior can still enable it manually:

~~~sh
cp docker-compose.override.yml.bak docker-compose.override.yml
~~~

This preserves flexibility for development without keeping override configuration as a first-class tracked file.

### DIFF
--- a/core/services/storages/cloud/minio.py
+++ b/core/services/storages/cloud/minio.py
@@ -21,10 +21,12 @@ class MinIO(CloudStorageServiceInterface):
         self.secure = settings.MINIO_SECURE
         self.client = Minio(endpoint=self.endpoint, access_key=self.access_key, secret_key=self.secret_key,
                             secure=self.secure)
-        # Separate client for presigned URL generation using the externally accessible endpoint.
+        # Dedicated client for presigned URL generation.
+        # Uses the external endpoint so the signature carries the publicly accessible host.
         # Region is set explicitly to avoid a network round-trip for region discovery
         # (the external endpoint is not reachable from inside the container).
-        self.external_client = Minio(
+        # All other operations use self.client (internal endpoint) since they run server-side.
+        self._presign_client = Minio(
             endpoint=self.external_endpoint, access_key=self.access_key, secret_key=self.secret_key,
             secure=self.secure, region=settings.MINIO_REGION
         ) if self.external_endpoint != self.endpoint else self.client
@@ -90,7 +92,7 @@ class MinIO(CloudStorageServiceInterface):
         Falls back to the internal client when no external endpoint is configured.
         """
         try:
-            return self.external_client.get_presigned_url(
+            return self._presign_client.get_presigned_url(
                 method='GET', bucket_name=self.bucket_name, object_name=file_path
             ) if file_path else None
         except S3Error as e:
--- a/core/services/storages/cloud/minio.py
+++ b/core/services/storages/cloud/minio.py
@@ -14,12 +14,20 @@ class MinIO(CloudStorageServiceInterface):
     def __init__(self):
         super().__init__()
         self.endpoint = settings.MINIO_ENDPOINT
+        self.external_endpoint = settings.MINIO_EXTERNAL_ENDPOINT or self.endpoint
         self.access_key = settings.MINIO_ACCESS_KEY
         self.secret_key = settings.MINIO_SECRET_KEY
         self.bucket_name = settings.MINIO_BUCKET_NAME
         self.secure = settings.MINIO_SECURE
         self.client = Minio(endpoint=self.endpoint, access_key=self.access_key, secret_key=self.secret_key,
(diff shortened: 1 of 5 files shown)


docker-compose.override.yml.bak
```diff
@@ -20,9 +33,17 @@ services:
...
+      - rustfs
     environment:
       - ENVIRONMENT=development
       - DEBUG=${DEBUG-TRUE}
+      - NO_LM=TRUE
+      - EXPORT_SERVICE=core.services.storages.cloud.minio.MinIO
+      - MINIO_ENDPOINT=rustfs:9000
+      - MINIO_ACCESS_KEY=${MINIO_ACCESS_KEY-rustfsadmin}
+      - MINIO_SECRET_KEY=${MINIO_SECRET_KEY-rustfsadmin}
+      - MINIO_BUCKET_NAME=${MINIO_BUCKET_NAME-oclapi2-dev}
```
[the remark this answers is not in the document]
filiperochalopes: Solved

filiperochalopes:
image

image

image

image

core/services/storages/cloud/minio.py
```diff
@@ -14,12 +14,20 @@ class MinIO(CloudStorageServiceInterface):
...
         self.access_key = settings.MINIO_ACCESS_KEY
         self.secret_key = settings.MINIO_SECRET_KEY
         self.bucket_name = settings.MINIO_BUCKET_NAME
         self.secure = settings.MINIO_SECURE
         self.client = Minio(endpoint=self.endpoint, access_key=self.access_key, secret_key=self.secret_key,
                             secure=self.secure)
+        # Separate client for presigned URL generation using the externally accessible endpoint.
+        # Region is set explicitly to avoid a network round-trip for region discovery
+        # (the external endpoint is not reachable from inside the container).
+        self.external_client = Minio(
```
snyaggarwal: there are usages of `self.client`; shouldn't they all move to `self.external_client`?

filiperochalopes: Good point on the naming — `external_client` was ambiguous. I renamed it to `_presign_client` to make the intent explicit: it is only used for presigned URL generation, so the signature carries the publicly accessible host. All other operations (`upload_file`, `exists`, `has_path`, `delete_objects`, etc.) intentionally continue using `self.client`, since they run server-side inside the container, where only `MINIO_ENDPOINT` is reachable through the Docker network.

If we move toward using a single external-style variable for everything, the idea is reasonable, but it is not the fastest path for developers because it does not take advantage of Docker networking and the ready-to-use RustFS service. In that setup, we also would not be able to use `localhost`, since it would not resolve externally. We would need to introduce a proxy plus an `/etc/hosts` entry, which is a much more complete setup for what is currently intended to be a simple development configuration.

In any case, if you already have MinIO or RustFS running as an externally reachable service on your machine, defining only `MINIO_ENDPOINT` is enough to configure both `self.client` and `self._presign_client`.

snyaggarwal (approved):

PR #853, opened 2026-04-15
OpenConceptLab/ocl_issues#1854 | References API can conditionally include resolved repo versions

### LINKED ISSUE #1854: Reference
References are the “instructions” to tell OCL what resources to put in your collection and how to do it. References are akin to FHIR's logical definition, sometimes known as the Compose statement of a [ValueSet](https://www.hl7.org/fhir/valueset.html). 

In OCL, references often point to either concepts or mappings; they can also point to something broader e.g. all concepts in a repository. You can have one reference per concept that you want in your collection, or you can have one reference that brings in multiple concepts or mappings. Additionally, any reference can either be an INCLUDE or an EXCLUDE reference, which determines whether the concept(s)/mapping(s) should be brought into the collection or if they should be specifically omitted from the collection.

References also can have "Cascade" parameters, which tell OCL to not only get a single concept, but also to traverse the associations (i.e. mappings or hierarchy) of a concept to bring in more concepts. For example, users will often add an entire Concept Set to their collection, which brings in the concept for the set, plus any question concepts within that set, plus any answer concepts underneath those questions, and so on.

References can sometimes point to a specific version of a repository, but they often simply point to one or more concepts/mappings themselves without specifying a version. In this case, version resolution happens as part of building a collection expansion. Best practices for modeling references can be found in this document: https://docs.google.com/document/d/1QvWX0H6TP1_yJSiNkE9xWIrYmn9opp6O0-qQp57m4zc/edit

snyaggarwal:


paynejd:
Nice, focused change — the `INCLUDE_*` opt-in pattern and the model-method extraction both match existing conventions. Two blockers before merge:

**1. Tests.** The conditional field toggle and the new `get_resolved_repo_versions_serialized()` both need coverage:
- `?includeResolvedRepoVersions=true` → field present with expected shape
- default → field omitted from response
- reference with only `system`, only `valueset`, both, and neither

**2. N×resolution performance.** With the flag on, every reference triggers `resolve_system_version` + `resolve_valueset_versions`, each calling the full `resolve_reference_expression` algorithm (URL registry lookups, namespace match, version pinning). A collection with 100+ references will do 100+ resolution chains per list request. Worth addressing before this ships — a few options, roughly by effort:

- **Request-scoped memoization** *(cheapest).* Many refs in a collection point to the same system. Cache by `(namespace, url, version)` across the request — the expansion builder already does exactly this at [core/collections/models.py:1363](https://github.com/OpenConceptLab/oclapi2/blob/master/core/collections/models.py#L1363). Lift that helper into the serializer.
- **Read from the expansion's M2M instead of re-resolving.** `Expansion.evaluated_source_versions` / `evaluated_collection_versions` / `explicit_*` already hold resolution results. Prefetch rather than recompute. This is also closer to what ocl_issues#1854 asks for (the expansion-dropdown requirement implies resolved repo should be *contextual to an expansion*, not re-derived from reference fields) — may warrant a follow-up ticket.
- **Fast path for the canonical-URL branch.** Batch via `Source.objects.filter(canonical_url__in=[...])` for refs that don't hit URL-registry logic, fall back per-row otherwise. More code paths to maintain — only worth it if memoization isn't enough.

My suggestion: add memoization here to unblock merge, and split the expansion-context question into a follow-up since it shapes the API contract.

paynejd (approved):
both comments addressed. lgtm

PR #857, opened 2026-04-20
OpenConceptLab/ocl_issues#2480 | clone operation everything or nothing

### LINKED ISSUE #2480: $clone operation everything or nothing
Problem:
Right now `$clone` operation saves/clones what it can and it may silently ignore errors with some `concepts` or `mappings`.

Solution:
1. `$clone` should be everything or nothing
2. It should return errors against each `concept` and `mapping` that failed

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_issues#2480

### DIFF
--- a/core/concepts/views.py
+++ b/core/concepts/views.py
@@ -50,6 +50,7 @@ from core.concepts.serializers import (
     ConceptVersionListSerializer, ConceptSummarySerializer, ConceptMinimalSerializer,
     ConceptChildrenSerializer, ConceptParentsSerializer, ConceptLookupListSerializer, ConceptChecksumSerializer)
 from core.mappings.serializers import MappingListSerializer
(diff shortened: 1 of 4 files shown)


paynejd:
Flagging a regression risk spotted during review:

This PR adds `except CloneError` handling to `SourceConceptsCloneView.post` in `core/sources/views.py` (L400-402), but `ConceptCloneView.post` in `core/concepts/views.py` (L471-474) also calls `Bundle.clone(...)` and doesn't catch `CloneError`.

Before this PR, a failed clone through that endpoint silently succeeded. After this PR, it will raise an uncaught `CloneError` and surface as HTTP 500 — a behavioral regression on that entry point.

Suggest mirroring the same `try/except CloneError` block so both `$clone` entry points return the structured 400 response consistently.

PR #858, opened 2026-04-23
OpenConceptLab/ocl_online#76 | superuser are exempt from throttling

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_online#76

### DIFF
--- a/core/common/tests.py
+++ b/core/common/tests.py
@@ -1347,7 +1347,7 @@ class ChecksumViewTest(OCLAPITestCase):
 }, clear=True)
 @override_settings(ENABLE_THROTTLING=True)
 class ThrottleUtilTest(OCLTestCase):
-    """Verify throttle selection for guest, standard, and core auth groups."""
+    """Verify throttle selection for guest, standard, core, and superuser auth groups."""
 
     def test_get_throttles_for_guest_user(self):
         throttles = ThrottleUtil.get_throttles_by_user_plan(AnonymousUser())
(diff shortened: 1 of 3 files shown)


paynejd (commented):
## Review Comments

### 1. Superuser check may only skip header reporting, not actual throttle enforcement

This middleware's `process_response` sets the `X-LimitRemaining-*` response headers — but the actual rate limiting (returning 429s) is handled by DRF's `UserRateThrottle.allow_request()` at the view level. If `ocladmin` is actually being rate-limited (not just seeing header values), this change alone won't fix it.

The superuser exemption likely also needs to go into `ThrottleUtil.get_throttles_by_user_plan()` and `get_match_throttles_by_user_plan()` in `core/common/throttling.py`:

```python
@staticmethod
def get_throttles_by_user_plan(user):
    if not settings.ENABLE_THROTTLING:
        return []
    if get(user, 'is_superuser'):
        return []
    # ... rest of logic
```

(Same for `get_match_throttles_by_user_plan`.)

Could you confirm where the actual 429s are coming from? If it's the DRF view-level throttles, this PR as-is would skip the headers but still enforce the limits.

### 2. No tests

A test verifying that a superuser request bypasses throttle logic would be valuable here, especially since the triage bot flagged this as moderate risk (incorrect throttling changes could affect system performance or security).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

paynejd (approved):
Previous review comments addressed: superuser exemption correctly moved into `ThrottleUtil.get_throttles_by_user_plan` / `get_match_throttles_by_user_plan`, which drives DRF view-level enforcement across all view modules (not just middleware headers). Tests cover both selectors. Ticket #76 intent fully addressed. LGTM.

PR #859, opened 2026-04-23
Use case: CIEL Markdown changelog using OCL engine

filiperochalopes:
Solves https://github.com/OpenConceptLab/ocl_issues/issues/2325


core/common/checksums.py
```diff
@@ -363,14 +443,31 @@ def process(self):  # pylint: disable=too-many-locals,too-many-branches,too-many
                         continue
                     traversed_concepts.add(concept_id)
                     concept_db_id = self.concepts_diff.get_db_id_for(key, concept_id)
-                    concept = Concept.objects.filter(id=concept_db_id).first()
+                    if is_enriched:
+                        concept = concepts_cache.get(concept_db_id)
+                    else:
+                        concept = Concept.objects.filter(id=concept_db_id).first()
```
snyaggarwal, line 472: dont think we need this

core/common/tasks.py
```diff
@@ -880,20 +880,30 @@ def generate_key(*args, **kwargs):
...
 
     from core.sources.models import Source
     version1 = Source.objects.get(uri=version1_uri)
     version2 = Source.objects.get(uri=version2_uri)
-    fn = Source.changelog if is_changelog else Source.compare
-    result = fn(version1, version2, verbosity)
+
+    if is_changelog:
+        result = Source.changelog(version1, version2, verbosity)
+        if format_type == 'markdown':
```
snyaggarwal, line 901: for markdown should I think `format_type = 'md'`
just following the convention

core/common/checksums.py
filiperochalopes: yes, still working on that

core/common/tasks.py
filiperochalopes: The format is `markdown`, `md` is just the extension. The other part of the code uses `markdown` as well. Which convention are we talking about?

core/common/checksums.py
filiperochalopes: Yeah, I also did not like it much initially, it made the reading a bit harder, but the split is intentional for performance.

For verbosity < 4, we only need the concept itself to read lightweight fields like display_name, so a simple targeted lookup is enough and avoids building an upfront cache for every diff item. For verbosity >= 4, we need much heavier data: names, descriptions, previous concept state, and related objects used to build the detailed changelog. In that path, using concepts_cache lets us batch-fetch all required current and previous concepts once, with prefetch_related('names', 'descriptions'), instead of triggering repeated queries inside the loop.

So the conditional avoids paying the cost of the enriched cache for normal verbosity levels, while still preventing N+1 queries when the enriched changelog is requested.

core/common/tasks.py
snyaggarwal: following the `json` convention.

Ideally the API should be `?format=<json/xml/md>`

Following RFC convention
Accept: application/json
Accept: application/xml

core/common/checksums.py
filiperochalopes: We can keep the simpler, no-cache approach with a single bulk request for better maintainability and readability — your call.

snyaggarwal: I think its ok in current shape, I missed the usage

snyaggarwal (approved):

PR #861, opened 2026-04-23
Improves https://github.com/OpenConceptLab/oclapi2/pull/852

filiperochalopes:
After some real-world testing, I noticed that this approach increases hierarchy processing and parallel import tasks exponentially. A better option would be to create a separate task for hierarchy processing only after concept import has finished, which would reduce the processing load.


filiperochalopes:
Fixed and tested

snyaggarwal (approved):

PR #862, opened 2026-04-24
OpenConceptLab/ocl_issues#2485 | API to get project configurations

### LINKED ISSUE #2485: Create a new Project from an existing project
**What this would mean:**
Given an existing project, 
When user wants to create a new project from this project
Then they should start on new project confiugration screen prefilled from prev project

**Tasks:**
- [x] Copy only configurations and not candidates/decisions/logs/input-file or any other project identities
- [x] This option should be available on Project listing and Project Home screen.

snyaggarwal:
## Linked Issue

Ref OpenConceptLab/ocl_issues#2485

### DIFF
--- a/core/map_projects/serializers.py
+++ b/core/map_projects/serializers.py
@@ -93,7 +93,16 @@ class MapProjectListSerializer(serializers.ModelSerializer):
         ]
 
 
-class MapProjectSerializer(serializers.ModelSerializer):
+class MapProjectConfigurationsSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = MapProject
+        fields = [
+            'id', 'url', 'algorithms', 'encoder_model', 'filters', 'include_retired', 'lookup_config',
+            'score_configuration', 'target_repo_url'
+        ]
+
+
+class MapProjectSerializer(MapProjectConfigurationsSerializer):
     created_by = CharField(source='created_by.username', read_only=True)
     updated_by = CharField(source='updated_by.username', read_only=True)
     owner = CharField(source='parent.mnemonic', read_only=True)
@@ -105,13 +114,12 @@ class MapProjectSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = MapProject
-        fields = [
-            'id', 'name', 'input_file_name',
-            'created_by', 'updated_by', 'created_at', 'updated_at', 'url', 'is_active',
+        fields = MapProjectConfigurationsSerializer.Meta.fields + [
+            'name', 'input_file_name',
+            'created_by', 'updated_by', 'created_at', 'updated_at', 'is_active',
             'owner', 'owner_type', 'owner_url', 'public_access',
-            'target_repo_url', 'summary', 'logs', 'include_retired',
-            'score_configuration', 'filters', 'candidates', 'algorithms', 'lookup_config', 'analysis',
-            'encoder_model'
+            'summary', 'logs', 'include_retired',
+            'candidates', 'analysis',
         ]
 
     def __init__(self, *args, **kwargs):
@@ -137,6 +145,7 @@ class MapProjectDetailSerializer(MapProjectSerializer):
     class Meta:
         model = MapProject
         fields = MapProjectSerializer.Meta.fields + ['file_url', 'matches', 'columns']
+
 
 class MapProjectLogsSerializer(serializers.ModelSerializer):
     class Meta:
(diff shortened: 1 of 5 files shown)


core/map_projects/serializers.py
```diff
@@ -93,7 +93,16 @@ class Meta:
         ]
 
 
-class MapProjectSerializer(serializers.ModelSerializer):
+class MapProjectConfigurationsSerializer(serializers.ModelSerializer):
```
paynejd, line 96: Once `CONFIGURATION_FIELDS` lives on the model (see top-level review comment), replace the hardcoded list here with `['id', 'url'] + MapProject.CONFIGURATION_FIELDS`. Also worth a one-line docstring on the class stating the rule (e.g. "Fields that define *how* a project matches — excluding identity, results, logs, and audit metadata. Used by the copy-project flow.") so the contract is discoverable.

Two field-scope questions worth confirming explicitly (in code or PR description):

- **`description`** is in `MapProjectCreateUpdateSerializer` but excluded here — copy won't carry the description forward. Intentional? (Arguably correct since the description likely refers to the original project, but worth being deliberate.)
- **`public_access`** is in the base serializer but excluded here — the copied project gets default access rather than inheriting. Likely correct, but same point: worth being deliberate.

core/map_projects/tests/tests.py
```diff
@@ -120,3 +121,35 @@ def test_put(self, upload_mock):
         self.assertEqual(len(response.data['columns']), 1)
         upload_mock.assert_called_once_with(
             key=f"map_projects/{response.data['id']}/input.csv", file_content=ANY)
+
+
+class MapProjectConfigurationsViewTest(MapProjectAbstractViewTest):
```
paynejd, line 126: Test asserts that included fields are correct, but doesn't assert the **ticket's actual contract**: that excluded fields are absent. Please add assertions that the response does NOT contain `candidates`, `logs`, `summary`, `analysis`, `input_file_name`, `matches`, `columns`. That's what guarantees we're not leaking project identity/results into a copy.

Also missing:

- A negative auth test (unauthorized user → 403/404). Other map-project views are protected by `CanEditConceptDictionary`; one assertion confirming this endpoint inherits that protection would lock it down.
- A 404 test for a non-existent project ID.

paynejd (commented):
Solid foundation — the inheritance refactor is the right move and the new endpoint slots cleanly into existing URL/view patterns. A few items to address before merge, mostly around making the "what is a configuration" rule explicit and tightening the test coverage to actually enforce the ticket's scope contract.

---

## Define `CONFIGURATION_FIELDS` as a class constant on the `MapProject` model

(Leaving this at the top level since `core/map_projects/models.py` isn't in this PR's diff.)

Right now "what counts as a project configuration" lives only as a hardcoded field list in `MapProjectConfigurationsSerializer`. There's no single source of truth — the model is flat, with config fields (`algorithms`, `filters`, `lookup_config`, …) interleaved with result/runtime fields (`matches`, `candidates`, `analysis`, …) and identity/audit fields. Whether a field is "configuration" is a human judgment that's currently undocumented.

Please add a class-level constant on `MapProject`, e.g.:

```python
class MapProject(BaseModel):
    ...
    CONFIGURATION_FIELDS = [
        'algorithms', 'encoder_model', 'filters', 'include_retired',
        'lookup_config', 'score_configuration', 'target_repo_url',
    ]
```

Then have `MapProjectConfigurationsSerializer.Meta.fields` reference it (e.g. `['id', 'url'] + MapProject.CONFIGURATION_FIELDS`). This:

1. Puts the definition next to the field declarations, where future contributors will see it when adding a new field.
2. Makes it reusable from `MapProjectCreateUpdateSerializer` and any future copy-related logic, so a new config field automatically flows everywhere instead of needing four serializers updated by hand.
3. Documents intent — anyone reading the model can see the configuration boundary without reverse-engineering it from one of five serializers.

core/map_projects/serializers.py
snyaggarwal: Ya `name`, `description` are kind of part of _identity_ of the project. We shouldn't copy.
Same goes for "public_access", it should be default.

core/map_projects/tests/tests.py
snyaggarwal: 403 tests are not needed since its inheriting for the other views which are tested.
404 tests are not needed because no code is written for that, its handled by django framework and URL constructs

PR #865, opened 2026-05-04
OpenConceptLab/ocl_issues#2494 | locale retired

### LINKED ISSUE #2494: Add retired fields to Concept Name model for OpenMRS compatibility (retired + retiredReason)
## User story
As a maintainer, I want the OCL data model to support retired metadata on Concept Names so version exports remain compatible with OpenMRS/CIEL Terminology Interface imports and do not lose historically-retired name data.

## Use case
Export a CIEL version and import it into OpenMRS (CIEL Terminology Interface), preserving Concept Name retirement status and reason as used by OpenMRS schema 1.11/OBS workflows.

## Requirements
- [ ] Extend the OCL Concept Name data model to include:
  - [ ] retired (boolean)
  - [ ] retiredReason (string, optional)
- [ ] Ensure version exports include these Concept Name fields for retrocompatibility with OpenMRS import.
- [ ] Address existing historical data:
  - [ ] Preserve the currently existing retired Concept Names in CIEL (not lose retroactive data).
  - [ ] Note: ~5,474 retired Concept Names exist today across versions.
- [ ] Update the OCL Module synchronization layer to support the new Concept Name fields when syncing to/from OpenMRS.
- [ ] Assess and document required changes to keep OpenMRS sync functional with the updated model.

## Acceptance criteria
- [ ] Concept Name supports retired (boolean) and optional retiredReason (string) in the OCL data model.
- [ ] Version exports include Concept Name retired metadata required for OpenMRS/CIEL Terminology Interface imports.
- [ ] Existing retired Concept Names are preserved (no loss of historical retired data).
- [ ] OCL Module sync is updated (or a concrete plan is documented) to handle the new fields without breaking OpenMRS synchronization.

snyaggarwal:
## Linked Issue

Ref OpenConceptLab/ocl_issues#2494

- `retired` flag on `concept.names` and `concept.descriptions`
- only not retired locales are searchable


paynejd (commented):
## Review Notes

**Overall:** Solid approach — soft-delete via `retired` flag, consistent filtering with `active_names`/`active_descriptions`, good partial indexes, and nice test coverage.

A few things worth checking:

1. **CodeQL flag on views.py:728** — the `or e` fallback in the `except ValidationError` block passes the raw exception to the Response. Should be `or str(e)` or `or e.messages` to avoid leaking stack trace info.

2. **GraphQL inconsistency** — `serialize_names` (queries.py:172) returns all names including retired, but `resolve_description` (queries.py:543) filters to `active_descriptions` only. Is that intentional?

3. **`all_names` property removed** — was anything else referencing it (exports, OCL Module)?

4. **Migration 0081** drops and recreates the `preferred_locale` index on ConceptName. With 20M+ rows that index rebuild will take a lock — worth scheduling during a maintenance window.

snyaggarwal:
1. Fixed
2. serialize_names returns all names, while resolve_description resolves to preferred one so retired needs to be excluded from here.
3. all_names was dead code
4. tried on local with staging dump and took around 60seconds for indexes migrations. In any case will run this with migration service.

filiperochalopes:
I noticed that `concept_name.retired_reason` (string) was not implemented. Was this an intentional architectural decision to keep the scope concise @paynejd ? 

Some terminologies use a controlled set of coded reasons, while CIEL uses an open string field. These values exist in the CIEL/OpenMRS data model and could be lost if not implemented. That said, I’m not sure yet what the actual impact of that loss would be. @askanter

paynejd (approved):
Approving — minor items only. Worth resolving before/after merge:

1. `retiredReason` not implemented — confirm with Filipe whether to add a field or wire user input into the existing comment field on PUT/DELETE.
2. Historical CIEL retired-name backfill — Andrew's suggestion to populate from current CIELAdmin state is unanswered on the ticket; current default=False silently sides with no-migration.
3. OCL Module sync layer update — file a follow-up ticket so it doesn't get lost.
4. Partial rename: `ConceptLabelListCreateView` was left as-is while sibling classes became `ConceptLocale*` — either finish or revert.
5. Migration 0081 adds 5 new indexes to `concept_descriptions` beyond what this ticket needs — fine to keep, but flag it for the migration service run.

PR #866, opened 2026-05-07
OpenConceptLab/ocl_issues#2383 | saving prompt template url

### LINKED ISSUE #2383: [OCL Mapper] Power User AI Assistant Drop-Down for Model & Prompt Template Selection
## Summary
Add a configuration drop-down to the OCL Mapper AI Assistant, visible only to users with the `core_user` role, that allows selection of:
1. **AI Model** - similar to current behavior, however the default selection may be different for each prompt
2. **Prompt Template** - populated by calling the new AI Assistant `/prompts/` endpoint with filter `action_type=match_recommend`; default value for this is where the `prompt_key` matches the `action_type`, so it's the same as the current default `match_recommend`

## Default / Fallback Behavior
Users without the required role are unaffected. The AI Assistant continues to use prompt template key `match_recommend` with action type `match_recommend` as it does today. This is also the default selection for eligible users who haven't made a choice.

## Notes
- Prompt templates sharing the same `action_type` are expected to share the same input/output schema, so templates returned for `match_recommend` are schema-compatible and safe to swap in.

snyaggarwal:
## Linked Issue

Refs: OpenConceptLab/ocl_issues#2383


paynejd (commented):
## Review

Small, focused PR — does what it says. Two items to address before merge:

### 1. Naming: reconsider `prompt_template_url`

The ticket discussion settled on storing the template **without the version** — so what we're actually persisting is just the prompt template key (e.g. `match-recommend`), not a full URI (`/prompts/match-recommend/v1/`) or URL. Storing a key in a field called `_url` is misleading.

Suggest renaming to `prompt_template_key` (or just `prompt_template`) to match what's actually stored. The full URI is only needed when capturing the AI Assistant response (which includes the version used), not for the project-level config. Better to get the name right now than do a rename migration later.

### 2. Add a test

PR #862 added `MapProjectConfigurationsViewTest` — please add an assertion there (or a new test) that `prompt_template_url` (or whatever it gets renamed to) round-trips through the configurations endpoint and is included in create/update.

---

Everything else looks clean. The `encoder_model` AlterField in the migration is a harmless Django artifact (settings reference baked as literal) — no action needed.

paynejd (approved):
Both prior review concerns addressed — field renamed to `prompt_template_key` and round-trip test added. LGTM.

PR #867, opened 2026-05-07
fix(importers): sort bulk import inputs by resource type

filiperochalopes:
## Problem

`BulkImportParallelRunner.make_parts()` currently groups only consecutive same-type resources into processing parts.

When a client submits an interleaved payload such as:

~~~text
Concept → Mapping → Concept → Mapping
~~~

each type transition creates a new single-item part. These tiny parts are then dispatched as independent Celery tasks and processed sequentially, effectively removing any parallelism benefits.

This behavior became evident during ICD-11 imports:

- ~37k concepts were processed efficiently in 2 parallel chunks of ~18k items each
- ~6k postcoordinated concepts interleaved with mappings degraded into ~1 item per task
- each task introduced an additional ~5s scheduling/processing gap
- resulting in an estimated ~30x slowdown for that import segment

## Solution

Introduce a stable sort of `input_list` by resource type priority at the beginning of `make_parts()` before grouping logic executes.

Proposed ordering:

~~~text
concept → mapping → reference
~~~

This guarantees that:

- all concepts become consecutive
- all mappings become consecutive
- all references become consecutive

regardless of the payload ordering provided by the client.

As a result, imports generate at most 3 large processing groups instead of potentially thousands of 1-item tasks.

## Why this is safe

The proposed change preserves current semantic behavior:

- Stable sort preserves the original relative ordering within the same resource type
  - parent concepts remain before children if originally ordered correctly
- `org/source/collection` resources are already extracted before this stage
- Mapping resolution queries concepts from the database, not from the transient input buffer
  - therefore processing all concepts before mappings remains semantically equivalent
- `DELETE` and `__action` handling inside `queue_tasks()` remains untouched
- `collect_concept_hierarchy_map()` is already type-filtered and order-independent

## Expected impact

Bulk imports containing interleaved resource types transition from:

~~~text
O(n) sequential single-item tasks
~~~

to:

~~~text
O(types) grouped parallel processing
~~~

This directly improves throughput for any client that does not pre-sort payloads before submission.

### DIFF
--- a/core/importers/models.py
+++ b/core/importers/models.py
@@ -1014,12 +1014,25 @@ class BulkImportParallelRunner(BaseImporter):  # pragma: no cover
             self.resource_distribution[data_type].append(data)
 
     def make_parts(self):
-        _type_order = {'concept': 0, 'mapping': 1, 'reference': 2}
+        # Reflects the dependency order: containers must exist before their contents,
+        # and versioning (snapshots) must happen after all content has been imported.
+        # Organization → Source/Collection → Concept → Mapping → Reference → Versions
+        # Any unknown type falls back to 8, placing it after all known types.
+        _type_order = {
+            'organization':       0,
+            'source':             1,
+            'collection':         2,
+            'concept':            3,
+            'mapping':            4,
+            'reference':          5,
+            'source version':     6,
+            'collection version': 7,
+        }
         self.input_list = sorted(
             self.input_list,
             key=lambda item: _type_order.get(
                 (item if isinstance(item, dict) else json.loads(item)).get('type', '').lower(),
-                0
+                8
             )
         )
         prev_line = None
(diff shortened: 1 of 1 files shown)


snyaggarwal:
@filiperochalopes this is a sample json
https://github.com/OpenConceptLab/oclapi2/blob/master/core/samples/sample_ocldev.json

This goes like:
Org -> Source -> Source Version (initial) -> Concepts -> Mappings -> Source Version (first release) -> Concepts

Its an expected behaviour from importer to import multiple versions from single file.

I think we would be better off if we take an explicit input to sort.

filiperochalopes:
Hmm, you’re right — I wasn’t aware of that use case, but it does make sense.

In that case, I think it’s better to close this PR since it’s not really a quick win. I’ll adjust my script to generate a better-structured input JSON instead, which should consequently improve bulk import performance with parallel processing.

PR #875, opened 2026-05-26
OpenConceptLab/ocl_issues#2519 | retired_reason in concept/mapping and concept locales

### LINKED ISSUE #2519: OCLOMRS: Support Concept Name Retirement (Voiding) and Name-Level Void Reasons
## User story
As a Terminology Manager using the OCL Module, I want the subscription/import process to recognize when specific names of a concept are retired (voided) and to preserve the reason for that retirement, so my OpenMRS dictionary remains synchronized with CIEL’s data model and no editorial context is lost.

## Use case
A user subscribes to a source (e.g., CIEL) where a concept has multiple names. In a new version, one of these names is marked as retired: true. OCL propagates the reason for this retirement via the update_comment at the concept version level. The OCL Module must parse this information during import to update the OpenMRS concept_name table, setting voided = 1 and populating void_reason.

## Requirements
- [ ] Support names[].retired in import logic
  - [ ] The importer in openmrs-module-openconceptlab must check the retired boolean for each item in names[].
  - [ ] If retired: true, the corresponding name in OpenMRS must be voided (concept_name.voided = 1).
- [ ] Parse name-level void reasons from update_comment
  - [ ] Implement a parser for the established pattern:
    - [ ] Pattern: Retired name reason (NAME_UUID): REASON_TEXT
  - [ ] Match NAME_UUID to the name’s external_id (or uuid, when applicable) to extract the specific reason.
  - [ ] Handle multiple retired names in a single update_comment (multiple occurrences of the pattern).
- [ ] Ensure data integrity in exports
  - [ ] Import must rely on the OCL version export file (bulk download).
  - [ ] Verify the export includes name-level retired flag and the concept-level update_comment.
- [ ] Prevent data loss
  - [ ] If a name is voided and a reason is found, set void_reason to the extracted reason text.
  - [ ] If a name is voided and no matching reason is found, set a generic void_reason (e.g., “Retired in OCL”).
- [ ] Keep concept active when only names are retired
  - [ ] If concept retired=false and name retired=true, only the name is voided; the concept remains active.

## Acceptance criteria
- [ ] Given a concept export includes:
  - [ ] names[].retired = true for a specific name, and
  - [ ] update_comment contains Retired name reason (external_id): reason,
  - [ ] When the OCL Module imports it,
  - [ ] Then the matching OpenMRS concept_name record is updated with:
    - [ ] voided = 1
    - [ ] void_reason = extracted reason
- [ ] Given a name marked retired=true but no matching reason in update_comment,
  - [ ] When imported,
  - [ ] Then the name is still voided=1 and void_reason is set to a generic value.
- [ ] Given concept retired=false but name retired=true,
  - [ ] When imported,
  - [ ] Then the concept remains active in OpenMRS and only the specific name is voided.

### Reference JSON (OCL API v2)

```json
{
  "id": "1482",
  "retired": false,
  "names": [
    {
      "uuid": "259770136",
      "name": "Estructura del sistema cardiovascular",
      "external_id": "82622700-fa57-4431-99c3-b54343c22635",
      "retired": true
    }
  ],
  "update_comment": "Retired name reason (82622700-fa57-4431-99c3-b54343c22635): Este termo é obsoleto"
}
```

Expected OpenMRS result for the matching concept_name:
- voided = 1
- void_reason = "Obsolete term replaced by ..."

### Example requests

```sh
curl -s -X GET \
  "https://api.staging.openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1482/19380208/" \
  -H "Authorization: Token ***" \
  -H "Accept: application/json" | jq
```

```json
{
  "type": "Concept",
  "uuid": "19380208",
  "id": "1482",
  "external_id": "1482AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
  "concept_class": "Anatomy",
  "datatype": "N/A",
  "display_name": "Cardiovascular system",
  "display_locale": "en",
  "names": [
    {
      "uuid": "259770133",
      "name": "Cardiovascular system",
      "external_id": "1705BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB",
      "type": "ConceptName",
      "locale": "en",
      "locale_preferred": true,
      "name_type": "FULLY_SPECIFIED",
      "checksum": "b7a4699102829a2ddcd394e1730978d3",
      "retired": false
    },
    {
      "uuid": "259770134",
      "name": "Système cardio-vasculaire",
      "external_id": "1706BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB",
      "type": "ConceptName",
      "locale": "fr",
      "locale_preferred": true,
      "name_type": "FULLY_SPECIFIED",
      "checksum": "104df2980c35da0c7333d015dc78e86c",
      "retired": false
    },
    {
      "uuid": "259770135",
      "name": "Sistema circulatorio",
      "external_id": "b5777f44-0685-4602-8e17-73143b81b1fc",
      "type": "ConceptName",
      "locale": "es",
      "locale_preferred": false,
      "name_type": null,
      "checksum": "163a3aba94c6496b86ddf56f0b524e7d",
      "retired": false
    },
    {
      "uuid": "259770136",
      "name": "Estructura del sistema cardiovascular",
      "external_id": "82622700-fa57-4431-99c3-b54343c22635",
      "type": "ConceptName",
      "locale": "es",
      "locale_preferred": false,
      "name_type": null,
      "checksum": "12fbb3b03354112f3bbedf16f7406afa",
      "retired": true
    },
    {
      "uuid": "259770137",
      "name": "Estructura del aparato cardiovascular",
      "external_id": "5b585fea-fd26-4207-bfd0-1ab91dce4cdb",
      "type": "ConceptName",
      "locale": "es",
      "locale_preferred": false,
      "name_type": null,
      "checksum": "1b8ad49c99e2c77b6500475c03a84d73",
      "retired": false
    },
    {
      "uuid": "259770138",
      "name": "Mfumo wa damu mwilini",
      "external_id": "1707BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB",
      "type": "ConceptName",
      "locale": "sw",
      "locale_preferred": true,
      "name_type": "FULLY_SPECIFIED",
      "checksum": "69cfac6b133262dbc78f1c8000b5e98a",
      "retired": false
    },
    {
      "uuid": "259770140",
      "name": "Sistema cardiovascular",
      "external_id": "244ce535-9ac1-4799-8a78-200fb96f7826",
      "type": "ConceptName",
      "locale": "es",
      "locale_preferred": false,
      "name_type": null,
      "checksum": "a19f38e4fe3f7813f9a3ab1d3ac7536d",
      "retired": false
    },
    {
      "uuid": "259770141",
      "name": "Aparato cardiovascular",
      "external_id": "6330ba4b-e4f0-4a75-8a2a-5906fb8d03bb",
      "type": "ConceptName",
      "locale": "es",
      "locale_preferred": false,
      "name_type": null,
      "checksum": "93ab423844eab9c7b11f5e276c58120a",
      "retired": false
    },
    {
      "uuid": "259770139",
      "name": "Estructura cardiovascular",
      "external_id": "a1811490-03fd-4c74-b700-735cf262f64d",
      "type": "ConceptName",
      "locale": "es",
      "locale_preferred": false,
      "name_type": "FULLY_SPECIFIED",
      "checksum": "892c9f51a00df17b8e4e559a95aab5cc",
      "retired": false
    }
  ],
  "descriptions": [],
  "extras": {},
  "retired": false,
  "source": "CIEL",
  "source_url": "/orgs/CIEL/sources/CIEL/",
  "owner": "CIEL",
  "owner_name": "CIEL",
  "owner_url": "/orgs/CIEL/",
  "version": "19380208",
  "created_on": "2026-05-14T13:29:51.970865Z",
  "updated_on": "2026-05-14T13:29:52.008412Z",
  "version_created_on": "2026-05-14T13:29:51.970865Z",
  "version_created_by": "contato@filipelopes.med.br",
  "update_comment": "Retired name reason (82622700-fa57-4431-99c3-b54343c22635): [Insert the reason here]",
  "is_latest_version": true,
  "locale": null,
  "url": "/orgs/CIEL/sources/CIEL/concepts/1482/",
  "owner_type": "Organization",
  "version_url": "/orgs/CIEL/sources/CIEL/concepts/1482/19380208/",
  "mappings": [],
  "previous_version_url": "/orgs/CIEL/sources/CIEL/concepts/1482/19380206/",
  "source_versions": [],
  "collection_versions": [],
  "versioned_object_id": 310484,
  "checksums": {
    "smart": "24a2612b89bf88a79f41418d97aa0bcc",
    "standard": "a58795bb57c4ef41c1cbcae83821208a"
  },
  "version_updated_on": "2026-05-14T13:29:52.008412Z",
  "version_updated_by": "contato@filipelopes.med.br",
  "latest_source_version": null
}
```

<details>
<summary>Notes from review (PR context)</summary>

- retiredReason not implemented: confirm whether to add a field or wire user input into existing comment field on PUT/DELETE.
- Historical CIEL retired-name backfill needs a decision (populate from current CIELAdmin state vs no migration).
- OCL Module sync layer update should be tracked explicitly (this issue).
- Reference: https://github.com/OpenConceptLab/oclapi2/pull/865#pullrequestreview-4263987528

</details>

snyaggarwal:
## Linked Issue

Ref OpenConceptLab/ocl_issues#2519

## Summary

- adds `retire_reason` in `concept`, `mapping`, `concept.names[]`, `concept.descriptions[]`
- `retire_reason` is optionally set when retiring resource and copied along with future updates on it if the resource is still retired
- `update_comment` on `concept` and `mapping` is auto set when retiring (no chnage)
- on reactiavte of resource, `retire_reason` is set to None


filiperochalopes (approved):
`concept description` and `mapping` were not in the initial plan, but it looks good for standardizing.

filiperochalopes:
@snyaggarwal please wait for @paynejd 's review

filiperochalopes:
@snyaggarwal might be useful to have codex reviewer implemented in your side as well https://developers.openai.com/codex/app/review to have it's feedback in your PR's

core/concepts/serializers.py
```diff
@@ -80,7 +80,14 @@ def create(self, validated_data, instance=None):  # pylint: disable=arguments-di
         locale.name = validated_data.get('name', locale.name)
         locale.locale = validated_data.get('locale', locale.locale)
         locale.locale_preferred = validated_data.get('locale_preferred', locale.locale_preferred)
+
+        retired = validated_data.get('retired', None)
+        if retired is True and not locale.retired:
+            locale.retire_reason = validated_data.get('retire_reason', None)
+        elif locale.retired and retired is False:
+            locale.retire_reason = None
         locale.retired = validated_data.get('retired', locale.retired)
```
paynejd: `retire_reason` is a managed attribute — admins should be able to update it any time (not only on the retire transition), and we shouldn't silently null it when they reactivate. Real use case: an admin keeps a running note across cycles ("retired 2026-05-01, reactivated 2026-06-15 because users still need this"). The UI can offer to clear it on reactivate, but the API shouldn't force it.

```suggestion
        if 'retire_reason' in validated_data:
            locale.retire_reason = validated_data.get('retire_reason')
        locale.retired = validated_data.get('retired', locale.retired)
```

core/concepts/views.py
```diff
@@ -745,6 +746,7 @@ def delete(self, request, *args, **kwargs):
             ]
             retired_locale = instance.clone()
             retired_locale.retired = True
+            retired_locale.retire_reason = request.data.get('retire_reason', None)
```
paynejd: Same managed-attribute rule — if the admin DELETEs the locale without sending `retire_reason`, we shouldn't clobber whatever value the prior version carried.

```suggestion
            if 'retire_reason' in request.data:
                retired_locale.retire_reason = request.data.get('retire_reason')
```

core/common/mixins.py
```diff
@@ -721,23 +730,24 @@ def parent_resource(self):
...
-        return self.__update_retire(False, comment or self.WAS_UNRETIRED, user)
+        return self.__update_retire(False, user, comment or self.WAS_UNRETIRED)
 
-    def __update_retire(self, retired, comment, user):
+    def __update_retire(self, retired, user, comment, reason=None):
         latest_version = self.get_latest_version() or self.get_last_version()
         new_version = latest_version.clone()
         new_version.retired = retired
         new_version.comment = comment
+        new_version.retire_reason = reason if retired else None
```
paynejd: Two problems with this line:

1. **Auto-clears `retire_reason` on every unretire.** Admins may want to retain prior reasons across cycles ("retired 2026-05-01, reactivated 2026-06-15 because users still need this"). The API shouldn't enforce a reset.
2. **Overrides the cloned value on retire when no `reason` is supplied** (e.g. plain `concept.retire(user)` from an importer that doesn't carry a reason) — should preserve what the prior version had unless explicitly given a new value.

Treat it as a normal managed field: only write when the caller actually passed something.

```suggestion
        if reason is not None:
            new_version.retire_reason = reason
```

core/concepts/models.py
```diff
@@ -584,6 +580,14 @@ def create_new_version_for(
...
         instance.comment = data.get('update_comment') or data.get('comment')
+        instance.retire_reason = data.get('retire_reason', instance.retire_reason)
+        is_retired = data.get('retired', None)
+        if is_retired is True and not instance.retired:
+            instance.retire_comment = instance.comment
+            instance.comment = instance.WAS_RETIRED
+        elif instance.retired and is_retired is False:
+            instance.retire_comment = None
+            instance.comment = instance.comment or instance.WAS_RETIRED
         instance.retired = data.get('retired', instance.retired)
```
paynejd, line 584: Several issues in this block:

1. **`instance.retire_comment = ...` is dead code.** No `retire_comment` field exists anywhere in the codebase — `grep -rn retire_comment core/` returns only the two write-only assignments in this PR. The attribute is set on a Python instance and thrown away. Drop it.
2. **PUT silently clobbers the user's `update_comment` on retire.** `instance.comment = WAS_RETIRED` overwrites whatever the caller sent in `update_comment`. The DELETE path (`Concept.retire(user, comment, reason)`) preserves the caller's comment — so the two endpoints now have divergent semantics. Don't clobber.
3. **Unretire fallback uses the wrong constant.** `instance.comment or WAS_RETIRED` resolves to `"Concept was retired"` on a reactivate-with-no-comment — the opposite of what just happened. Should be `WAS_UNRETIRED` if any default is needed.
4. Per the design discussion, `retire_reason` is a managed attribute — admins can update it any time, not only when toggling `retired`. The existing `instance.retire_reason = data.get('retire_reason', instance.retire_reason)` is correct (honor payload, otherwise preserve) — keep it, drop the rest of the transition gymnastics.

```suggestion
        instance.comment = data.get('update_comment') or data.get('comment')
        instance.retire_reason = data.get('retire_reason', instance.retire_reason)
        instance.retired = data.get('retired', instance.retired)
```

core/mappings/models.py
```diff
@@ -403,10 +397,18 @@ def create_new_version_for(cls, instance, data, user):
...
         instance.sort_weight = data.get('sort_weight', instance.sort_weight)
+        instance.retire_reason = data.get('retire_reason', instance.retire_reason)
+        is_retired = data.get('retired', None)
+        if is_retired is True and not instance.retired:
+            instance.retire_comment = instance.comment
+            instance.comment = instance.WAS_RETIRED
+        elif instance.retired and is_retired is False:
+            instance.retire_comment = None
+            instance.comment = instance.comment or instance.WAS_RETIRED
+        instance.retired = data.get('retired', instance.retired)
```
paynejd: Same issues as the matching block in `core/concepts/models.py` — dead `retire_comment` assignments, PUT clobbers user-supplied `update_comment` on retire (inconsistent with the DELETE path), and the unretire fallback uses the wrong constant. Collapse to honoring the payload directly:

```suggestion
        instance.comment = data.get('update_comment') or data.get('comment')
        instance.mnemonic = data.get('mnemonic', instance.mnemonic)
        instance.map_type = data.get('map_type', instance.map_type)
        instance.sort_weight = data.get('sort_weight', instance.sort_weight)
        instance.retire_reason = data.get('retire_reason', instance.retire_reason)
        instance.retired = data.get('retired', instance.retired)
```

paynejd (changes requested):
Approach matches what we landed on in #2519 — `retire_reason` as a new managed field on concept and name (+ description and mapping for consistency). The refactor lifting shared fields into `SourceChildMixin` is a nice cleanup.

## Blocker — bulk export & detail responses don't expose name/description `retire_reason`

The whole ticket exists so the OCL Module can read `names[].retire_reason` out of the bulk-export ZIP and populate OpenMRS `concept_name.void_reason`. But the serializers used to render names/descriptions inside concept detail and `ConceptVersionExportSerializer` (`LocalizedNameSerializer` and `LocalizedDescriptionSerializer`, both untouched by this PR) don't list `retire_reason` in their `fields` tuples, so the field exists on the model and accepts writes but never appears in any GET or export. Current tests pass because they assert against model attributes (`concept.retired_names.first().retire_reason`) rather than serialized output.

Please update both:

**[core/concepts/serializers.py:26-29](https://github.com/OpenConceptLab/oclapi2/blob/master/core/concepts/serializers.py#L26-L29)** — `LocalizedNameSerializer.Meta.fields`:
```python
        fields = (
            'uuid', 'name', 'external_id', 'type', 'locale', 'locale_preferred', 'name_type', 'checksum',
            'retired', 'retire_reason'
        )
```

**[core/concepts/serializers.py:45-48](https://github.com/OpenConceptLab/oclapi2/blob/master/core/concepts/serializers.py#L45-L48)** — `LocalizedDescriptionSerializer.Meta.fields`:
```python
        fields = (
            'uuid', 'description', 'external_id', 'type', 'locale', 'locale_preferred', 'description_type', 'checksum',
            'retired', 'retire_reason'
        )
```

And please add an integration test that GETs a concept (and exercises the export serializer) and asserts both `names[].retire_reason` and `descriptions[].retire_reason` appear in the response.

## Treat `retire_reason` as a managed attribute — no auto-clear on reactivate, updatable any time

See inline suggestions on `__update_retire`, `ConceptLocaleSerializer.create`, the locale DELETE in `views.py`, and both `create_new_version_for` methods. The field is something admins curate, not something the API resets implicitly. Real use case: admin keeps a running note like *"retired 2026-05-01, reactivated 2026-06-15 because users still need it."* The UI can encourage clearing on reactivate, but the API shouldn't enforce it. Admins should also be able to PATCH `retire_reason` at any time without changing `retired`.

## Bulk import

Mechanically works — `retire_reason` is in `ConceptImporter.allowed_fields` and `MappingImporter.allowed_fields`, and nested `names[]`/`descriptions[]` flow through `AbstractLocalizedText._build()` which splats the dict. But please add an importer test that imports a concept with `names[].retire_reason` populated and asserts it lands on the persisted name row.

## Smaller cleanups (inline)

- Drop dead `instance.retire_comment` assignments — no such field exists anywhere (`grep -rn retire_comment core/` returns only the two write-only assignments in this PR).
- Fix the unretire fallback that resolves to `WAS_RETIRED` ("Concept was retired") instead of `WAS_UNRETIRED` on a reactivate-with-no-comment.
- Reconcile the PUT-clobbers-`update_comment` behavior with the DELETE path which preserves the user's comment.

## Note on scope

The `retired` boolean on `concept.description` is pre-existing — it's on the `AbstractLocalizedText` parent class — and this PR adds `retire_reason` to the same parent, so both `ConceptName` and `ConceptDescription` inherit it. ✅

PR #876, opened 2026-05-29
AutomatchRun model + endpoints + project-scoped authz (ocl_online#109)

paynejd:
Implements **OpenConceptLab/ocl_online#109** — the **oclapi2 slice** of the retrospective process attribution design ([ocl_online#105](https://github.com/OpenConceptLab/ocl_online/issues/105) §A). Per #105's repo-sliced decomposition, this covers both oclapi2 work items: the `AutomatchRun` entity + REST surface, **and** forwarding the attribution headers from the analytics emitter.

## Part 1 — `AutomatchRun` entity (Phase 2)

`AutomatchRun(BaseModel)` in `core/map_projects/models.py` (`db_table='automatch_runs'`), with the locked field set from #105 §A, `BigAutoField` id (OQ2), and the two `Meta` indexes `[map_project, -started_at]` / `[completion_status]`. Migration `0038`.

| Method | Path | Purpose |
|---|---|---|
| `POST` | `/…/map-projects/<id>/auto-match-runs/` | create at run start, returns `id` |
| `GET` | `/…/map-projects/<id>/auto-match-runs/` | list a project's runs |
| `GET` | `/auto-match-runs/<id>/` | fetch a single run |
| `PATCH` | `/auto-match-runs/<id>/` | update progress / completion (PATCH-only; PUT → 405) |

Locked decisions honored:
- **Project/ownership-scoped authz on every endpoint (OQ2)** — anchored on the parent `MapProject` via `CanEditConceptDictionary` (consistent with how MapProject's own read views are gated); the sequential run id is not a security boundary. A user cannot read/patch/list another project's runs.
- **Re-run semantics (OQ3)** — a failed-row re-run is a **new** run with `parent_run` set + `trigger_source='ui-rerun-row'`; `intended_rows` = the failed-row count; `config_snapshot` = config at retry time. The parent's failure snapshot is never mutated. A cross-project `parent_run` is rejected (400).
- **Immutable run-start snapshot** — `PATCH` is restricted to lifecycle fields (`completed_rows`, `failed_rows`, `completion_status`, `completed_at`); reaching a terminal status auto-stamps `completed_at`. `trigger_source` / `completion_status` validated against the #105 vocabularies (app-level, no DB CHECK).
- `started_by`, `client_ip`, `client_user_agent` captured server-side.

## Part 2 — forward the attribution headers (analytics emitter)

This is the second oclapi2 work item in #105's per-repo breakdown (*"add `request_source` + `event_metadata` to … paths that call `api_transactions`"*), surfaced during review.

oclapi2 records `api_transactions` by having `AnalyticsEventEmitter` POST a payload to ocl-analytics-api's `/api-transactions/` receiver. `_safe_request_headers` forwards only an **allow-list** of request headers, and the `X-OCL-*` attribution headers were absent — so on every oclapi2-served call (`$match`, `$resolveReference`, concept-detail GET) they were stripped, and the analytics-api parser ([#111](https://github.com/OpenConceptLab/ocl_online/issues/111)) would see nothing (rows stay `unknown`/`{}`). The ai-assistant `$invoke` path is unaffected.

Added the five attribution headers to `ALLOWED_REQUEST_HEADERS` so they survive into `payload["request"]["headers"]` (raw `HTTP_X_OCL_*` keys, like the existing forwarded headers): `request_source`, `event_metadata`, `request_idempotency_key`, `prompt_template_key`, `prompt_template_version`. This is a prerequisite for #111 to attribute oclapi2-served traffic — **#111 should parse these keys from the forwarded payload.**

## Scope notes
- Entity-only for the rest: the Mapper UI consumes the endpoints in Phase 5 (oclmap), deferred until PR3-C (oclmap#35) + #2461 land.
- No model/middleware coupling between the two parts — the emitter change is global request middleware, independent of the entity.

## Tests & checks
- 24 tests pass: `core.map_projects` (run CRUD, rerun linkage + parent immutability, vocabulary validation, PUT→405, cross-project / non-member authz) + `core.services` (attribution headers survive the allow-list; allow-list stays strict).
- `pylint` 10.00/10 on all changed modules; `makemigrations --check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


core/services/tests/tests.py
```diff
@@ -0,0 +1,35 @@
...
+class AnalyticsEventEmitterTest(OCLTestCase):
+    def test_attribution_headers_are_forwarded(self):
+        # The #105 attribution headers must survive the request-header allow-list,
+        # since the emitter forwards only allow-listed headers into the
+        # api-transactions payload it POSTs to ocl-analytics-api. If they were
+        # dropped, the analytics-api parser (ocl_online#111) would receive nothing
+        # and every oclapi2-served row would stay request_source='unknown'/{}.
+        request = RequestFactory().get(
+            '/concepts/',
+            HTTP_X_OCL_REQUEST_SOURCE='automatch',
```
snyaggarwal, line 16: all of these headers must be defined in `settings.py#CORS_ALLOW_HEADERS`

snyaggarwal: Added these

snyaggarwal (approved):

PR #879, opened 2026-06-12
Restrict hard deletion of versioned concepts (solves https://github.com/OpenConceptLab/ocl_issues/issues/2570)

### LINKED ISSUE #2570: Allow source editors to hard delete unpublished concepts
## User story
As a CIEL Lab curator, I want to permanently delete a newly created concept that exists only in OCL HEAD so I can correct creation mistakes before the concept is published.

## Use case
CIEL Lab users can create concepts in OCL HEAD. If a concept was created by mistake and has never been included in a release or draft source version, it should be removable by users who already have write access to the source. Currently, only administrators can perform this hard delete, which limits concept management from CIEL Lab.

## Requirements
- [ ] Allow `DELETE ?hardDelete=true` for authenticated users with write permission on the source.
- [ ] Allow the operation only when the concept exists exclusively in HEAD.
  - [ ] Check every row belonging to the concept within the same source.
  - [ ] Block deletion if any current or historical concept version belongs to a non-HEAD source version.
  - [ ] Treat both released and draft source versions as non-HEAD versions.
- [ ] Return `409 Conflict` when a source editor attempts to hard delete a concept associated with a non-HEAD source version.
- [ ] Preserve the existing unrestricted hard delete behavior for administrators.
- [ ] Keep `async=true`, `db=true`, and individual concept-version deletion restricted to administrators.
- [ ] Keep regular `DELETE` behavior unchanged when `hardDelete=true` is not provided.
- [ ] Prevent concurrent versioning and deletion from bypassing the eligibility validation.
- [ ] Update the source concept count after a successful hard delete.
- [ ] Do not require a database migration or public schema change.

## Acceptance criteria
- [ ] Given a concept exists only in HEAD, when the owner of a private user source requests a hard delete, then the concept is permanently deleted.
- [ ] Given a concept exists only in HEAD, when a member of the owning organization requests a hard delete, then the concept is permanently deleted.
- [ ] Given a concept exists only in HEAD and the source has public edit access, when an authenticated user requests a hard delete, then the concept is permanently deleted.
- [ ] Given a user does not have write access to a private source, when they request a hard delete, then the request is rejected and the concept remains unchanged.
- [ ] Given an anonymous user, when they request a hard delete, then the request is rejected and the concept remains unchanged.
- [ ] Given any version of a concept belongs to a released source version, when a non-administrator requests a hard delete, then the API returns `409 Conflict`.
- [ ] Given any version of a concept belongs to a draft source version, when a non-administrator requests a hard delete, then the API returns `409 Conflict`.
- [ ] Given a historical concept version was published but the latest version exists only in HEAD, when a non-administrator requests a hard delete, then the API returns `409 Conflict`.
- [ ] Given a hard delete is blocked, then concepts, versions, names, descriptions, mappings, and source-version associations remain unchanged.
- [ ] Given a concept belongs to a non-HEAD source version, when an administrator requests a hard delete, then the existing administrative behavior is preserved.
- [ ] Given a source editor requests hard delete with `async=true` or `db=true`, then the request is rejected.
- [ ] Given a source editor requests hard delete for an individual concept version, then the request is rejected.
- [ ] Given a source editor sends a regular `DELETE` without `hardDelete=true`, then the concept is retired instead of permanently deleted.
- [ ] Given two concepts share the same mnemonic in different sources, when one is eligible for hard delete, then version history from the other source does not block or affect its deletion.
- [ ] Given an eligible concept is deleted, then the source concept count is updated.

## More details (collapsible)
<details>
<summary>Context and integrity considerations</summary>

### Context

CIEL Lab relies on OCL HEAD for concept creation and editing. Newly created concepts may need to be removed before publication when they were created by mistake or are no longer needed.

The current OCL API authorization restricts hard delete to administrators, even when the requesting user already has write permission on the source and the concept has never been published.

### Eligibility definition

A concept is considered HEAD-only when none of the concept rows identified by its source and versioned object identity is associated with a source version other than HEAD.

The validation must not rely only on the latest concept row because an earlier concept version may already belong to a release.

### Security

Eligibility validation and deletion must be performed atomically. The relevant concept rows should be locked during validation to prevent a concurrent source-version operation from making an ineligible concept appear eligible.

</details>

filiperochalopes:
## Summary

Allows users with write permission on a source to hard delete concepts that have never belonged to a draft or released source version.

This removes the administrator dependency that previously limited unpublished concept management from CIEL Lab.

## Authorization decisions

- User source owners may delete eligible concepts.
- Organization members may delete eligible concepts from organization sources.
- Usuário autenticado exclui conceito de source com acesso público de edição.
- Users without write access and anonymous users remain blocked.
- Administrators retain unrestricted hard delete access.
- `async=true`, `db=true`, and individual version deletion remain administrator-only.

## Eligibility rule

A non-administrator may hard delete a concept only when none of its current or historical rows is associated with a source version other than HEAD.

Draft and released versions both make the concept ineligible. Blocked requests return `409 Conflict`.

## Complete history deletion

The endpoint deletes the versioned concept root. Its version rows reference that root using `on_delete=CASCADE`, so the operation removes every HEAD-only concept version and its dependent names, descriptions, and associations.

A regression test creates four edits, producing six database rows including the root and initial version, and verifies that all six are removed.

## Integrity considerations

- All source-scoped concept rows are locked with `select_for_update`.
- Eligibility validation and deletion occur in one transaction.
- Blocked deletion preserves concepts, versions, names, descriptions, mappings, and source-version associations.
- Source concept counts are updated after successful deletion.

## Tests

- 16 new authorization, history, cascade, and integrity tests: passing.
- 4 existing delete and retire regression tests: passing.
- 20 relevant tests total: passing.
- Pylint: `10.00/10`.
- Python compilation and `git diff --check`: passing.

## Files changed

- `core/concepts/constants.py`
- `core/concepts/models.py`
- `core/concepts/views.py`
- `core/integration_tests/tests_concepts.py`

No database migration or public schema change is required.

### DIFF
--- a/core/concepts/models.py
+++ b/core/concepts/models.py
@@ -21,7 +21,6 @@ from core.concepts.constants import CONCEPT_TYPE, LOCALES_FULLY_SPECIFIED, LOCAL
     MAX_NAMES_LIMIT, MAX_DESCRIPTIONS_LIMIT
 from core.concepts.mixins import ConceptValidationMixin
 from core.services.storages.postgres import PostgresQL
-from core.tasks.models import Task
 
 
 class AbstractLocalizedText(ChecksumModel):
@@ -901,13 +900,12 @@ class Concept(ConceptValidationMixin, SourceChildMixin, VersionedModel):  # pyli
 
     def belongs_to_non_head_source_version(self):
         """Return whether any version of this source-scoped concept was snapshotted."""
+        # Snapshots link the `is_latest_version=True` row (see Source.seed_concepts), not the
+        # versioned-object row this method is usually called on, so aggregate across all versions.
         versioned_object_id = self.versioned_object_id or self.id
-        concept_ids = Concept.objects.filter(
-            parent_id=self.parent_id,
-            versioned_object_id=versioned_object_id,
-        ).values_list('id', flat=True)
         return Concept.sources.through.objects.filter(
-            concept_id__in=concept_ids,
+            concept__parent_id=self.parent_id,
+            concept__versioned_object_id=versioned_object_id,
         ).exclude(source__version=HEAD).exists()
 
     def has_pending_source_version_seed(self):
@@ -919,19 +917,9 @@ class Concept(ConceptValidationMixin, SourceChildMixin, VersionedModel):  # pyli
         if not concept_created_at:
             return False
 
-        source = self.parent
-        source_version_ids = source.__class__.objects.filter(
-            mnemonic=source.mnemonic,
-            organization_id=source.organization_id,
-            user_id=source.user_id,
-            created_at__gte=concept_created_at,
-        ).exclude(version=HEAD).values_list('id', flat=True)
-
-        for source_version_id in source_version_ids:
-            seed_task = Task.find(
-                name__iendswith='seed_children_to_new_version',
-                args__contains=['source', source_version_id],
-            )
+        source_versions = self.parent.versions.exclude(version=HEAD).filter(created_at__gte=concept_created_at)
+        for source_version in source_versions:
+            seed_task = source_version.get_seed_new_version_task()
             if seed_task and seed_task.state != SUCCESS:
                 return True
         return False
--- a/core/concepts/models.py
+++ b/core/concepts/models.py
@@ -1,3 +1,4 @@
+from celery.states import SUCCESS
 from django.conf import settings
 from django.contrib.postgres.indexes import HashIndex
 from django.core.exceptions import ValidationError
@@ -20,6 +21,7 @@ from core.concepts.constants import CONCEPT_TYPE, LOCALES_FULLY_SPECIFIED, LOCAL
     MAX_NAMES_LIMIT, MAX_DESCRIPTIONS_LIMIT
 from core.concepts.mixins import ConceptValidationMixin
 from core.services.storages.postgres import PostgresQL
+from core.tasks.models import Task
 
 
 class AbstractLocalizedText(ChecksumModel):
@@ -907,6 +909,32 @@ class Concept(ConceptValidationMixin, SourceChildMixin, VersionedModel):  # pyli
         return Concept.sources.through.objects.filter(
             concept_id__in=concept_ids,
         ).exclude(source__version=HEAD).exists()
+
+    def has_pending_source_version_seed(self):
+        """Return whether an unfinished source snapshot may include this concept."""
+        versioned_object_id = self.versioned_object_id or self.id
+        concept_created_at = Concept.objects.filter(id=versioned_object_id).values_list(
+            'created_at', flat=True
+        ).first()
+        if not concept_created_at:
+            return False
+
+        source = self.parent
+        source_version_ids = source.__class__.objects.filter(
+            mnemonic=source.mnemonic,
+            organization_id=source.organization_id,
+            user_id=source.user_id,
+            created_at__gte=concept_created_at,
+        ).exclude(version=HEAD).values_list('id', flat=True)
+
+        for source_version_id in source_version_ids:
+            seed_task = Task.find(
+                name__iendswith='seed_children_to_new_version',
+                args__contains=['source', source_version_id],
+            )
+            if seed_task and seed_task.state != SUCCESS:
+                return True
+        return False
 
     def get_source_version_before_creation(self):
         return self.sources.exclude(version=HEAD).filter(
--- a/core/concepts/models.py
+++ b/core/concepts/models.py
@@ -1,4 +1,3 @@
-from celery.states import SUCCESS
 from django.conf import settings
 from django.contrib.postgres.indexes import HashIndex
 from django.core.exceptions import ValidationError
@@ -897,33 +896,6 @@ class Concept(ConceptValidationMixin, SourceChildMixin, VersionedModel):  # pyli
     @property
     def latest_source_version(self):
         return self.sources.exclude(version=HEAD).order_by('-created_at').first()
-
-    def belongs_to_non_head_source_version(self):
-        versioned_object_id = self.versioned_object_id or self.id
-        versions = Concept.objects.filter(
-            parent_id=self.parent_id,
-            versioned_object_id=versioned_object_id,
-        ).prefetch_related('sources')
-        return any(
-            any(source.version != HEAD for source in version.sources.all())
-            for version in versions
-        )
-
-    def has_pending_source_version_seed(self):
-        """Return whether an unfinished source snapshot may include this concept."""
-        versioned_object_id = self.versioned_object_id or self.id
-        concept_created_at = Concept.objects.filter(id=versioned_object_id).values_list(
-            'created_at', flat=True
-        ).first()
-        if not concept_created_at:
-            return False
-
-        source_versions = self.parent.versions.exclude(version=HEAD).filter(created_at__gte=concept_created_at)
-        for source_version in source_versions:
-            seed_task = source_version.get_seed_new_version_task()
-            if seed_task and seed_task.state != SUCCESS:
-                return True
-        return False
 
     def get_source_version_before_creation(self):
         return self.sources.exclude(version=HEAD).filter(
--- a/core/concepts/models.py
+++ b/core/concepts/models.py
@@ -899,14 +899,15 @@ class Concept(ConceptValidationMixin, SourceChildMixin, VersionedModel):  # pyli
         return self.sources.exclude(version=HEAD).order_by('-created_at').first()
 
     def belongs_to_non_head_source_version(self):
-        """Return whether any version of this source-scoped concept was snapshotted."""
-        # Snapshots link the `is_latest_version=True` row (see Source.seed_concepts), not the
-        # versioned-object row this method is usually called on, so aggregate across all versions.
         versioned_object_id = self.versioned_object_id or self.id
-        return Concept.sources.through.objects.filter(
-            concept__parent_id=self.parent_id,
-            concept__versioned_object_id=versioned_object_id,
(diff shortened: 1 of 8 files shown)


filiperochalopes:
This solves https://github.com/OpenConceptLab/ocl_issues/issues/2570 @snyaggarwal @jamlung-ri

core/concepts/models.py
```diff
@@ -897,6 +899,43 @@ def is_existing_in_parent(self):
...
         return self.sources.exclude(version=HEAD).order_by('-created_at').first()
 
+    def belongs_to_non_head_source_version(self):
+        """Return whether any version of this source-scoped concept was snapshotted."""
+        versioned_object_id = self.versioned_object_id or self.id
+        concept_ids = Concept.objects.filter(
+            parent_id=self.parent_id,
+            versioned_object_id=versioned_object_id,
+        ).values_list('id', flat=True)
+        return Concept.sources.through.objects.filter(
```
snyaggarwal: this can cover both `Concept` and `ConceptVersion` by:
`self.sources.exclude(version=HEAD).exists()`

core/concepts/models.py
```diff
@@ -897,6 +899,43 @@ def is_existing_in_parent(self):
...
+        """Return whether an unfinished source snapshot may include this concept."""
+        versioned_object_id = self.versioned_object_id or self.id
+        concept_created_at = Concept.objects.filter(id=versioned_object_id).values_list(
+            'created_at', flat=True
+        ).first()
+        if not concept_created_at:
+            return False
+
+        source = self.parent
+        source_version_ids = source.__class__.objects.filter(
```
snyaggarwal: versions exclude HEAD gte concept created
`versions = self.parent.versions.exclude(version=HEAD).filter(created_at__gte=concept_created_at).values_list('id', flat=True)`

core/concepts/models.py
```diff
@@ -897,6 +899,43 @@ def is_existing_in_parent(self):
...
+            mnemonic=source.mnemonic,
+            organization_id=source.organization_id,
+            user_id=source.user_id,
+            created_at__gte=concept_created_at,
+        ).exclude(version=HEAD).values_list('id', flat=True)
+
+        for source_version_id in source_version_ids:
```
snyaggarwal:
```
for version in versions.only('id'):
  task = version.get_seed_new_version_task()
  if task and task.state != SUCCESS:
    return True
```

core/concepts/views.py
```diff
@@ -344,7 +347,10 @@ def get_permissions(self):
         if self.request.method in ['GET']:
             return [CanViewParentDictionary(), ]
 
-        if self.request.method == 'DELETE' and self.is_hard_delete_requested():
+        if (
+                self.request.method == 'DELETE' and self.is_hard_delete_requested() and
+                (self.is_async_requested() or self.is_db_delete_requested())
+        ):
```
snyaggarwal, line 356: this still doesnt include repo editor

core/concepts/models.py
filiperochalopes: The versioned-object row resolved here (`id == versioned_object_id`) is never `is_latest_version`, and `Source.seed_concepts` only snapshots the `is_latest_version=True` rows — so `self.sources.exclude(version=HEAD)` on the versioned object is always empty even when a version of the concept was released. We need to aggregate across all versions sharing `versioned_object_id`. Kept that, but simplified it to a single `through`-table query. 👍

core/concepts/models.py
filiperochalopes: Done — switched to `self.parent.versions.exclude(version=HEAD).filter(created_at__gte=concept_created_at)`.

core/concepts/models.py
filiperochalopes: Done — now using the existing `source_version.get_seed_new_version_task()` and checking state != SUCCESS. Also dropped the now-unused Task import.

core/concepts/views.py
filiperochalopes: Good catch — opened async hard delete to repo editors: only `db=true` (the raw bypass) stays admin-only now. The HEAD-only safety check runs before the async task is queued, so a non-admin still gets a 409 for a snapshotted concept. Updated the tests accordingly.

core/concepts/models.py
snyaggarwal: Concept.sources (or concept_sources) many to many relation ideally should only have source_id and concept_id where concept_id is the concept version and not concept (versioned_object) itself.

Thats why `self.sources` is best and it covers both Concept and interim versions as well

filiperochalopes: Applied the prefetch_related + self.sources approach.

The hard-delete operation still works at the concept entity/root level: deleting the versioned-object row lets CASCADE remove all related concept versions. Because of that, the guard still needs to answer the entity-level question:

“Has any version of this concept entity been snapshotted into a released source version?”

The updated implementation keeps that behavior, but uses self.sources on each version after prefetching all sources in batch:

~~~mermaid
flowchart TD
    Root["Concept root / versioned-object row<br/>id = versioned_object_id"] -->|"hard delete target"| Delete["Delete root row"]
    Delete -->|"CASCADE"| V1["Concept version v1"]
    Delete -->|"CASCADE"| V2["Concept version v2"]
    Delete -->|"CASCADE"| V3["Concept version v3"]

    SV1["Released Source Version SV1<br/>version != HEAD"] -->|"M2M snapshot"| V1
    HEAD["HEAD Source"] -. "working state" .-> Root

    Guard["Guard before hard delete"] --> Q1["Query 1:<br/>fetch all Concept versions<br/>with same versioned_object_id"]
    Q1 --> Q2["Query 2:<br/>prefetch_related('sources')<br/>loads sources for all versions in batch"]
    Q2 --> Py["Python check:<br/>for each version,<br/>check version.sources from prefetch cache"]
    Py --> Check{"Any source.version != HEAD?"}

    Check -->|"yes"| Block["Block hard delete"]
    Check -->|"no"| Allow["Allow root delete"]
    Allow --> Delete

    style Root stroke-width:3px
    style Guard stroke-width:3px
    style Block stroke-width:3px
    style Delete stroke-width:3px
~~~

So instead of checking only self.sources on the versioned-object row, the method now:

1. fetches all concept versions sharing the same versioned_object_id;
2. prefetches their sources in one batch;
3. checks source.version != HEAD in Python using the prefetched cache.

That keeps the same deletion rule: hard delete is allowed only if none of the concept versions are linked to a non-HEAD source version.

The permission behavior is unchanged. The existing guards in _db_hard_delete and _hard_delete still apply only for non-admin users, and check_object_permissions is untouched. The change is limited to the internal implementation of belongs_to_non_head_source_version().

One side note: the name sources is still a bit confusing here because Source is also a domain entity in the project. In this context, concept.sources behaves more like “source versions/snapshots that include this concept version” rather than the owning Source itself. I think our knowledge base would benefit from a short ubiquitous-language glossary for terms like Source, SourceVersion, HEAD, versioned object, concept version, and snapshot.

core/concepts/models.py
```diff
@@ -897,6 +898,33 @@ def is_existing_in_parent(self):
     def latest_source_version(self):
         return self.sources.exclude(version=HEAD).order_by('-created_at').first()
 
+    def belongs_to_non_head_source_version(self):
```
snyaggarwal: Can we consider Mapping in this PR as well? If yes, then:

this should move to `SourceChildMixin` so that Mapping can benefit for that as well.
same goes for `has_pending_source_version_seed`

core/concepts/models.py
```diff
@@ -897,6 +898,33 @@ def is_existing_in_parent(self):
...
+        versioned_object_id = self.versioned_object_id or self.id
+        versions = Concept.objects.filter(
+            parent_id=self.parent_id,
+            versioned_object_id=versioned_object_id,
+        ).prefetch_related('sources')
+        return any(
```
snyaggarwal: tried benchmarking against this logic

```
    def belongs_to_non_head_source_version(self):  # b2
        version_ids = self.__class__.objects.filter(
            parent_id=self.parent_id,
            versioned_object_id=self.versioned_object_id,
        ).values('id')

        return self.__class__.sources.through.objects.filter(
            concept_id__in=version_ids,
        ).exclude(
            source__version=HEAD,
        ).exists()
```

```
from timeit import repeat
b1_times = repeat(lambda: concept.b1(), repeat=5, number=100)
b2_times = repeat(lambda: concept.b2(), repeat=5, number=100) # b2 is above function

print("b1:", min(b1_times), b1_times)
b1: 0.7121121671516448 [1.4687205429654568, 0.7872442088555545, 0.7121121671516448, 0.7124972508754581, 0.7156632510013878]

print("b2:", min(b2_times), b2_times)
b2: 0.04778150003403425 [0.05238304194062948, 0.04903175006620586, 0.048412667121738195, 0.04778150003403425, 0.04811920807696879]
```

Tested this on Staging data for CIEL concept with 69 versions

core/concepts/models.py
filiperochalopes: Mapping was not my prior intention, but it's a great idea!

Both `belongs_to_non_head_source_version` and `has_pending_source_version_seed` now live on `SourceChildMixin`, so `Mapping` gets them too. Made them generic (`self.__class__`, through-table FK derived from `self._meta.model_name`).

core/concepts/models.py
filiperochalopes: Good catch — adopted the `b2` approach. It now queries the M2M through table directly with an `.exists()` instead of the prefetch + Python `any(...)` loop. Kept it generic for `Concept`/`Mapping` via the through model's FK name. Thanks for benchmarking! 🙏

snyaggarwal (approved):

PR #881, opened 2026-06-16
OpenConceptLab/ocl_issues#2477 | repo external exports

### LINKED ISSUE #2477: Add external exports support for OCL versions (upload/delete to S3-like bucket + auditable downloads)
## User story
As a release maintainer, I want to attach external export artifacts (e.g., sql.zip) to an OCL version so multiple CIEL export variants can be distributed from the version export UI with auditable downloads.

## Use case
After a version is created, upload one or more external export files (e.g., SQL exports for multiple OpenMRS versions) and make them available as selectable download options when exporting that version.

## Requirements
- [ ] Extend the export feature to support “external exports” attached to a specific OCL version.
- [ ] In the export UI, provide additional export options beyond the default export (e.g., sql.zip variants).
- [ ] Make download clicks auditable (record each download event).
- [ ] Implement file management endpoints for version-linked artifacts:
  - [ ] POST upload an external export file associated to a version (stored in S3-like bucket)
  - [ ] DELETE remove an external export file associated to a version (remove from S3-like bucket)
- [ ] Uploaded files must be stored in the S3-like bucket and retrievable by the export UI.

## Acceptance criteria
- [ ] Given a version, an external export file can be uploaded via a POST endpoint and becomes available as an export option for that version.
- [ ] Given an uploaded external export file, it can be removed via a DELETE endpoint and is no longer available for download.
- [ ] Given a user downloads any export option, a download audit record is created.

snyaggarwal:
## Linked Issue

Refs OpenConceptLab/ocl_issues#2477

OCL Web2 PR: https://github.com/OpenConceptLab/oclweb2/pull/36


core/common/models.py
```diff
@@ -1002,6 +1005,11 @@ def get_version_export_path(self, suffix='*'):
 
         return path
 
+    def get_external_export_path(self, key, filename):
```
filiperochalopes, line 1009: Simple sanitization using .replace() for slashes might be insufficient and risky.

1. **Path Traversal**: It doesn't handle '..' sequences, which can lead to unpredictable path resolution.
2. **OS Compatibility**: It preserves characters like ':', '*', '<', '>', which are invalid on Windows systems.
3. **Storage/URL Issues**: Spaces and special characters are kept, which often break S3 signed URLs and Content-Disposition headers during download.

**Recommendation**: Use django.utils.text.get_valid_filename(filename) for a robust, industry-standard solution.

```py
from django.utils.text import get_valid_filename

def get_external_export_path(self, key, filename):
    base_path = self.get_version_export_path(suffix=None).rstrip(".")
    safe_filename = get_valid_filename(filename)
    return f"{base_path}/external/{key}_{safe_filename}"
```

core/repos/serializers.py
```diff
@@ -2,10 +2,19 @@
...
 from core.sources.models import Source
 from core.sources.serializers import SourceListSerializer
 
 
+class RepoExternalExportSerializer(serializers.ModelSerializer):
+    url = serializers.CharField(source='uri', read_only=True)
+
+    class Meta:
+        model = RepoExternalExport
+        fields = ('key', 'description', 'url', 'created_at', 'updated_at', 'file_path')
```
filiperochalopes: Is `file_path` necessary here? Maybe remove it for security reasons, preventing disclosure of filesystem paths

filiperochalopes (changes requested):
Was the download audit intentionally left out?

Do we currently have any way to audit which users downloaded files through AWS? This audit trail is needed for some decisions @askanter needs to make regarding which versions should be kept and which users are still using those versions.

Also, locally I’ve had some issues where the 302 redirect did not always happen. I’m currently using a local patch to cover redirect validation to the signed bucket path.

It would be useful to add an integration test ensuring that the download GET works correctly and results in a 302 redirect.

snyaggarwal:
@filiperochalopes 
ocl-analytics tracks all requests.
GET /api-transactions/?operation_type=external_export&request_method=GET&ocl_resource=/orgs/CIEL/sources/CIEL/

Raised PR in analytics for `external_export` operation -- https://github.com/OpenConceptLab/ocl-analytics-api/pull/75

snyaggarwal:
@filiperochalopes already added [integration test](https://github.com/OpenConceptLab/oclapi2/blob/issues%232477/core/integration_tests/tests_sources.py#L1307) for redirect behaviour

filiperochalopes:
> @filiperochalopes ocl-analytics tracks all requests. GET /api-transactions/?operation_type=external_export&request_method=GET&ocl_resource=/orgs/CIEL/sources/CIEL/
> 
> Raised PR in analytics for `external_export` operation -- https://github.com/OpenConceptLab/ocl-analytics-api/pull/75

image

I might not have access to this repo

core/repos/serializers.py
snyaggarwal: for now `file_path` is used in oclweb2 to download the file with same name as the upload.

core/repos/mixins.py
```diff
@@ -0,0 +1,79 @@
...
+from core.repos.serializers import RepoExternalExportSerializer
+
+
+class RepoExternalExportMixin:
+    permission_classes = (CanViewConceptDictionaryVersion, IsAuthenticated)
+
+    def get_object(self):
+        queryset = self.get_queryset()
+        if 'version' not in self.kwargs:
+            queryset = queryset.filter(released=True)
```
paynejd: `filter(released=True).first()` is non-deterministic for the `latest/export/<key>/` route: neither `Source` nor `Collection` declares a `Meta.ordering`, so with multiple released versions `.first()` can return any of them. The standard `ConceptContainerExportMixin.get_object` filters on the `is_latest_version` flag (exactly one row) — align with it:

```suggestion
            queryset = queryset.filter(is_latest_version=True)
```

core/sources/views.py
```diff
@@ -559,6 +560,10 @@ def handle_export_version(self):
             return status.HTTP_409_CONFLICT
 
 
+class SourceVersionExternalExportView(RepoExternalExportMixin, SourceVersionBaseView):
+    permission_classes = (CanViewConceptDictionary, IsAuthenticated)
```
paynejd: The mixin already defines `permission_classes` (the version-scoped `CanViewConceptDictionaryVersion`, matching `ConceptContainerExportMixin`); this override shadows it with the non-version-scoped variant, so the mixin's copy is dead. Drop the override and inherit the mixin's:

```suggestion
class SourceVersionExternalExportView(RepoExternalExportMixin, SourceVersionBaseView):
    pass
```
This switches GET to the version-scoped permission (same as the core export view); please confirm the external-export tests stay green.

core/collections/views.py
```diff
@@ -1184,6 +1185,10 @@ def handle_export_version(self):
             return status.HTTP_409_CONFLICT
 
 
+class CollectionVersionExternalExportView(RepoExternalExportMixin, CollectionVersionBaseView):
+    permission_classes = (CanViewConceptDictionary, IsAuthenticated)
```
paynejd: Same as the source view — this override shadows the mixin's version-scoped `permission_classes`. Inherit it instead:

```suggestion
class CollectionVersionExternalExportView(RepoExternalExportMixin, CollectionVersionBaseView):
    pass
```

core/repos/serializers.py
paynejd: `file_path` exposes the raw S3 key (bucket layout + owner mnemonics) to every authenticated viewer. The frontend only needs it to name the downloaded file — expose a derived `filename` and drop the internal path:

```suggestion
class RepoExternalExportSerializer(serializers.ModelSerializer):
    url = serializers.CharField(source='uri', read_only=True)
    filename = serializers.SerializerMethodField()

    class Meta:
        model = RepoExternalExport
        fields = ('key', 'description', 'url', 'created_at', 'updated_at', 'filename')

    @staticmethod
    def get_filename(obj):
        return obj.file_path.split('/')[-1] if obj.file_path else None
```
Paired with the frontend change in OpenConceptLab/oclweb2#36.

paynejd (commented):
Reviewed for code quality + intent against #2477. The core design is clean — `RepoExternalExport` as a generic, repo-shared model attached via `GenericRelation` on `ConceptContainerModel` mirrors the existing `ConceptContainerExportMixin` conventions well, path-traversal is handled with a test, and coverage is solid across sources + collections. A few inline suggestions below; two non-blocking notes:

**Auditable downloads (AC):** confirmed the intended mechanism is the analytics request-logging pipeline — every download is an authenticated GET that lands in `api_transactions` — which is what we discussed and is acceptable. @snyaggarwal please verify a download is actually **queryable from the analytics service with sufficient metadata** (who / which repo+version / which export key). The GET returns a 302 to a signed URL and there's no test asserting the event is captured, so a quick confirmation that the transaction log carries enough to audit a download would close this out.

**Non-blocking:**
- **N+1:** `get_external_exports` runs one query per version; on the version *list* endpoint with `includeExternalExports=true` (which the UI sends) that's a query per row. A `prefetch_related('external_exports')` on that path removes it.
- **Atomicity:** upload/delete touch S3 then the DB with no rollback, so a partial failure can orphan an S3 object or leave a dangling row. Fine for a low-frequency admin action — worth a follow-up only if it ever runs unattended.

PR #886, opened 2026-06-24
OpenConceptLab/ocl_issues#2591 | caching lookups for mapping import

### LINKED ISSUE #2591: Speed up bulk import path
## Speed up bulk import path: dedupe redundant concept/source lookups

Worker concurrency and DB connection headroom are **intentionally not** levers here — concurrency is capped per-request by design for fair usage across simultaneous imports, and RDS has connection headroom to spare.

### Root cause
For each mapping, two independent code paths re-resolve the same `from_concept`/`to_concept`/source from URLs, with no caching across items in the same import run:

1. `MappingImporter.get_queryset()` (`core/importers/models.py`) resolves `from_concept`/`to_concept` via `Concept.objects.filter(uri=...).first()` and a `to_source` existence check, just to check if the mapping already exists.
2. `Mapping.persist_new()` → `populate_fields_from_relations()` (`core/mappings/models.py`) **re-resolves the same four things from scratch** moments later, when actually creating/updating the mapping.
3. Neither caches across mappings, even though many mappings in a typical source reference the same handful of concepts/sources (e.g. a shared target vocabulary).

Net: a single mapping does a dozen+ DB round trips, several of them exact repeats — either of its own earlier lookup or of another item's lookup for the same concept/source. At 350k mappings this compounds into millions of avoidable queries.

### Fix implemented
- Added a `cache` dict threaded through `BulkImportInline` → `ConceptImporter`/`MappingImporter` → `Mapping.persist_new()`/`create_new_version_for()` → `populate_fields_from_relations()`, scoped to one import sub-task (chunk).
- Cached: parent `Source` lookup by (owner, source) — shared across all concepts and mappings in a chunk; `Concept`-by-URI lookups (two separate namespaces, since `get_queryset()`'s and `populate_fields_from_relations()`'s query semantics differ slightly and aren't safe to unify); `Source.resolve_reference_expression()` results.
- Also throttled the per-item `Task.save()` progress write (`BulkImportInline.notify_progress()`) from once-per-item to once every 2 seconds (plus a forced final write), since it was issuing ~410k synchronous UPDATEs purely for progress tracking. Low risk, likely minutes of savings rather than a major driver.
- Files touched: `core/importers/models.py`, `core/mappings/models.py`.

### Benchmark
Added `tools/benchmark_bulk_import.py` — runs `BulkImportInline` against a sample inside a rolled-back transaction, reporting wall time and DB query count (via `CaptureQueriesContext`, with the 9000-entry query-log cap raised).

Measured on `core/samples/pepfar_datim_moh_fy19.json` (413 rows: 178 concepts, 222 mappings), same dev DB state both runs (old code via `git stash`, same data):

| | Baseline | With caching |
|---|---|---|
| DB queries | 24,515 | 22,696 (**-7.4%**) |
| Wall time | 14.55s | ~15.0s (noise at this size) |

Cache instrumentation confirmed correctness of the dedup (1 source lookup shared across all 413 items instead of 413 separate; 177 unique concept lookups instead of 600+ redundant ones).

### Open question / follow-up needed
The 7.4% query reduction is real but modest on this small sample — per-mapping fixed costs (versioning, `set_checksums`, `.sources.set()`, `update_mappings_count`) dominate over the concept/source-resolution savings this change targets. The win should scale better at production scale (350k mappings, high reuse of a shared target vocabulary, where unique-concept count stays flat while mapping count grows), but this hasn't been confirmed against a dataset of that size — **needs validation on a large/production-representative source before/after to get a real number.**

### Out of scope (confirmed with stakeholder)
- Worker concurrency / `parallel` param tuning (intentional fair-usage design)
- Postgres connection headroom (confirmed sufficient on RDS)
- S3/`import_type` chunked-upload path
- Supervisor in-memory polling/wait problem (`wait_till_tasks_alive`)
- Indexing strategy (already deferred/batched in 5000-item chunks)

### Verification / test plan
- [ ] Run existing importer/mapping test suites (`core/importers/tests.py`, `core/mappings/tests.py`)
- [ ] Re-run `tools/benchmark_bulk_import.py` against a large/production-representative source (ideally close to 60k concepts / 350k mappings) to confirm the dedup scales as expected
- [ ] Confirm `created`/`updated`/`failed`/`unchanged` counts are unchanged before/after on the same input
- [ ] Double-check cache correctness for in-chunk ordering edge cases: a mapping created earlier in a chunk that's referenced by a later mapping in the same chunk must not hit a stale "not found" cache entry from before it existed; same for concept versioning (re-fetch after `create_new_version_for`)

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_issues#2591

### DIFF
--- a/core/importers/models.py
+++ b/core/importers/models.py
@@ -109,12 +109,13 @@ class BaseResourceImporter:
     mandatory_fields = set()
     allowed_fields = []
 
-    def __init__(self, data, user, update_if_exists=False):
+    def __init__(self, data, user, update_if_exists=False, cache=None):
         self.user = user
         self.data = data
         self.update_if_exists = update_if_exists
         self.queryset = None
         self.index_resources = False
+        self.cache = cache if cache is not None else {}
 
     @classmethod
     def can_handle(cls, obj):
@@ -126,6 +127,15 @@ class BaseResourceImporter:
 
     def get(self, attr, default_value=None):
         return self.data.get(attr, default_value)
+
+    def get_cached_parent_source(self):
+        cache = self.cache.setdefault('source_by_owner', {})
+        key = (self.get_owner_type_filter(), self.get('owner'), self.get('source'))
+        if key not in cache:
+            cache[key] = Source.objects.filter(
+                **{self.get_owner_type_filter(): self.get('owner')}, mnemonic=self.get('source'), version=HEAD
+            ).first()
+        return cache[key]
 
     def parse(self):
         self.data = self.get_filter_allowed_fields()
@@ -416,8 +426,8 @@ class ConceptImporter(BaseResourceImporter):
     def get_resource_type():
         return 'Concept'
 
-    def __init__(self, data, user, update_if_exists, skip_hierarchy_tasks=False):
-        super().__init__(data, user, update_if_exists)
+    def __init__(self, data, user, update_if_exists, skip_hierarchy_tasks=False, cache=None):
+        super().__init__(data, user, update_if_exists, cache=cache)
         self.skip_hierarchy_tasks = skip_hierarchy_tasks
         self.version = False
         self.instance = None
@@ -439,9 +449,7 @@ class ConceptImporter(BaseResourceImporter):
         return self.queryset
 
     def parse(self):
-        source = Source.objects.filter(
-            **{self.get_owner_type_filter(): self.get('owner')}, mnemonic=self.get('source'), version=HEAD
-        ).first()
+        source = self.get_cached_parent_source()
         super().parse()
         self.data['parent'] = source
         self.data['mnemonic'] = str(self.data.pop('id', ''))
@@ -527,13 +535,25 @@ class MappingImporter(BaseResourceImporter):
     def get_resource_type():
         return 'Mapping'
 
-    def __init__(self, data, user, update_if_exists):
-        super().__init__(data, user, update_if_exists)
+    def __init__(self, data, user, update_if_exists, cache=None):
+        super().__init__(data, user, update_if_exists, cache=cache)
         self.version = False
         self.instance = None
 
     def exists(self):
         return self.get_queryset().exists()
+
+    def get_cached_versioned_concept_by_uri(self, uri):
+        cache = self.cache.setdefault('concept_versioned_by_uri', {})
+        if uri not in cache:
+            cache[uri] = Concept.objects.filter(id=F('versioned_object_id'), uri=uri).first()
+        return cache[uri]
+
+    def get_cached_source_exists_by_uri(self, uri):
+        cache = self.cache.setdefault('source_exists_by_uri', {})
+        if uri not in cache:
+            cache[uri] = Source.objects.filter(uri=uri).exists()
+        return cache[uri]
 
     def get_queryset(self):  # pylint: disable=too-many-branches
         if self.queryset:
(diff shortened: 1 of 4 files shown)


paynejd (commented):
Solid, well-scoped, low-risk performance refactor — backward compatible (every new param is `cache=None`; the other callers in `concepts/models.py`, `mappings/serializers.py`, `concept_maps/serializers.py` keep the old behavior), honest about its own limits, and aligned with existing patterns. Good restraint in keeping the two concept-lookup namespaces separate (`concept_versioned_by_uri` vs `concept_by_expr`) — they have genuinely different query semantics and unifying them would be wrong. The `notify_progress` throttle is a clean independent win, and `force=True` preserves the final summary.

A few notes from review, most-important first.

### 1. Negative-result caching + mid-chunk creation — please document the invariant it relies on

The caches store *misses* (`None`, `False`, dict fallback, `(None, …)`). If a concept/source is created during a run and an earlier item already cached its absence, a later item gets a stale miss.

Tracing it: in the **production parallel path this is safe** — `make_parts()` splits concepts and mappings into separate parts → separate chunks → separate caches, so a mapping chunk never creates the concepts it resolves, and a concept chunk's parent source already exists. The window only exists in the **deprecated `BulkImportInlineView`** / direct interleaved `BulkImportInline` calls.

So: low real-world risk, but it rests on the invariant "a chunk is single-resource-type," which isn't stated anywhere near the cache. A one-line comment at the cache site (and/or on `self.cache`) would keep a future chunking change from silently reintroducing staleness. This is the same edge case called out in the ticket's verification checklist.

### 2. A cache-correctness regression test would lock this down cheaply

Existing suites exercise the new signatures (good — confirms backward compat), but nothing asserts cache *correctness*: identical created/updated/unchanged counts with vs. without cache, plus the staleness case above (a mapping referencing a concept defined earlier in the same inline run). Given staleness is the one real risk, this is worth adding before merge — the ticket's verification checklist is currently all unchecked.

### 3. The benchmark belongs in the repo — make it a tracked, reusable artifact

The ticket describes `tools/benchmark_bulk_import.py`, but the PR is 2 files and doesn't include it. Beyond just committing it: this is exactly the kind of thing worth turning into a **reusable, automated artifact** rather than a one-off local script —
- commit it so the prod-scale before/after is reproducible by anyone;
- consider wiring it as a CI integration test (e.g. nightly / on importer changes) that runs `BulkImportInline` against a fixed sample in a rolled-back transaction and asserts on **DB query count** (via `CaptureQueriesContext`) and result counts;
- track the query-count number over time so future regressions in the import path surface as a failing/trending check, not as a production incident.

That also turns the still-open "needs validation at 350k-mapping scale" question into a standing, owned signal instead of a manual TODO.

### Minor
- **Memory:** `get_cached_versioned_concept_by_uri` retains the full `Concept` instance for the chunk lifetime but only `versioned_object_id` is used downstream — `.only('versioned_object_id')` (or `values_list`) trims this for large mapping chunks. (The `populate` cache genuinely needs the full object; leave it.)
- The `# pragma: no cover` moved onto the `self.task.summary = {` assignment, leaving the new throttle branches uncovered — relates to (2).

Net: clean and safe for the production path. I'd merge after confirming the benchmark's home and ideally adding the cache-correctness test. The honest caveat stands — the real win still needs a prod-scale before/after, which is the best argument for making the benchmark a tracked CI artifact.

PR #888, opened 2026-07-02
OpenConceptLab/ocl_issues#2599 | Performance improvements for Expansion concepts/mappings indexing

### LINKED ISSUE #2599: Speed up Expansion indexing by using partial doc update in ES
### Problem:
Expansion indexing is slow
Expansion1: 2,841 concepts + 12,545 mappings - ~194s
Expansion2: 3,073 concepts + 15,006 mappings - ~184s

### Tasks/Solution:
- [x] Collection url/version/mnemonic to be build using expansion URL, avoiding join
- [x] Use bulk queries in Concept/Mapping documents.py to collect collections fields.
- [x] Use ES Partial doc updates on Concepts/Mappings for collection/expansion values 

### Results (scirpt: tools/benchmark_batch_index.py)
Expansion1: 2,841 concepts + 12,545 mappings - ~194s~ 4.7s (42x)
Expansion2: 3,073 concepts + 15,006 mappings - ~184s~ 3.9s (47x)

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_issues#2599

## Summary

Collection | Resources | Full re-index | Partial update | Speedup
-- | -- | -- | -- | --
gigantic-collection/1 | 2,841 concepts + 12,545 mappings | 194s | 4.7s | 42x
PEPFAR MER_FY22/v1.0 | 3,073 concepts + 15,006 mappings | 184s | 3.9s | 47x

### DIFF
--- a/core/collections/models.py
+++ b/core/collections/models.py
@@ -1359,8 +1359,6 @@ class Expansion(BaseResourceModel):
     def delete_references(self, references):
         refs, _ = self.to_ref_list_separated(references)
 
-        index_concepts = []
-        index_mappings = []
         any_ref_with_resources = False
         for reference in refs:
             concepts = reference.concepts
@@ -1368,7 +1366,6 @@ class Expansion(BaseResourceModel):
                 any_ref_with_resources = True
                 resources_updated = list(concepts.values_list('versioned_object_id', flat=True))
                 filters = {'versioned_object_id__in': resources_updated}
-                index_concepts = [*index_concepts, *resources_updated]
                 self.concepts.set(self.concepts.exclude(**filters))
                 batch_index_resources.apply_async(('concept', filters), queue='indexing', permanent=False)
             mappings = reference.mappings
@@ -1376,11 +1373,8 @@ class Expansion(BaseResourceModel):
                 any_ref_with_resources = True
                 resources_updated = list(mappings.values_list('versioned_object_id', flat=True))
                 filters = {'versioned_object_id__in': resources_updated}
-                index_mappings = [*index_mappings, *resources_updated]
                 self.mappings.set(self.mappings.exclude(**filters))
                 batch_index_resources.apply_async(('mapping', filters), queue='indexing', permanent=False)
-
-        self.index_resources(index_concepts, index_mappings)
 
         if any_ref_with_resources:
             removed_reference_ids = [ref.id for ref in self.to_ref_list(references)]
@@ -1538,10 +1532,15 @@ class Expansion(BaseResourceModel):
             concepts, mappings = get_ref_results(reference)
             concepts_updated = self.__exclude_resources(reference, self.concepts, concepts, Concept)
             mappings_updated = self.__exclude_resources(reference, self.mappings, mappings, Mapping)
-            if index and concepts_updated is not False:
-                index_concepts = [*index_concepts, *concepts_updated.values_list('versioned_object_id', flat=True)]
-            if index and mappings_updated is not False:
-                index_mappings = [*index_mappings, *mappings_updated.values_list('versioned_object_id', flat=True)]
+            if index:
+                if concepts_updated is not False:
+                    filters = {
+                        'versioned_object_id__in': list(concepts_updated.values_list('versioned_object_id', flat=True))}
+                    batch_index_resources.apply_async(('concept', filters), queue='indexing', permanent=False)
+                if mappings_updated is not False:
+                    filters = {
+                        'versioned_object_id__in': list(mappings_updated.values_list('versioned_object_id', flat=True))}
+                    batch_index_resources.apply_async(('mapping', filters), queue='indexing', permanent=False)
 
         self.explicit_collection_versions.add(*compact(explicit_valueset_versions))
         self.explicit_source_versions.add(*compact(explicit_system_versions))
--- a/core/collections/models.py
+++ b/core/collections/models.py
@@ -30,7 +30,8 @@ from core.common.tasks import seed_children_to_expansion, batch_index_resources,
 from core.common.utils import drop_version, to_owner_uri, generate_temp_version, es_id_in, \
     get_resource_class_from_resource_name, to_snake_case, \
     es_to_pks, batch_qs, split_list_by_condition, decode_string, is_canonical_uri, encode_string, \
-    get_truthy_values, get_falsy_values, get_current_authorized_user, to_camel_case
+    get_truthy_values, get_falsy_values, get_current_authorized_user, to_camel_case, to_parent_uri, \
+    to_parent_kwargs_from_uri
 from core.concepts.constants import LOCALES_FULLY_SPECIFIED
 from core.concepts.models import Concept
 from core.mappings.models import Mapping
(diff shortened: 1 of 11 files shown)


paynejd (commented):
The ~40x speedup is legitimate — partial field-append via Painless script vs. re-serializing the whole concept/mapping document. And the URL-derivation / partial-update approach matches the tasks in #2599. But I found two correctness regressions the benchmark and tests don't cover (both removal-related and value-equivalence), plus one edge note. Details inline.

core/collections/models.py
```diff
@@ -1244,6 +1246,59 @@ def active_mappings(self):
     def owner_url(self):
         return to_owner_uri(self.uri)
 
+    @property
+    def collection_version_url(self):
+        return to_parent_uri(self.uri)
```
paynejd: **Bug: `collection_url` gets indexed as the *expansion* URI, not the collection-version URI.**

`to_parent_uri` only truncates at `/concepts/` or `/mappings/` — never `/expansions/`. An expansion URI (`/orgs/o/collections/c/v1.0/expansions/e1/`) contains neither, so it's returned unchanged. That value then feeds the `collection_url` ES field via **both** `_get_resources_index_collection_fields` (fast path) and the rewritten `prepare()` in `concepts/documents.py` and `mappings/documents.py`. The old code read `collection_version__uri` from the DB (`/orgs/o/collections/c/v1.0/`).

`collection_url` is a `filterable`/`facet` KeywordField (exact match), so filters on the collection-version URL (e.g. the one built in `common/views.py:509`) will stop matching every doc re-indexed after this change.

`to_parent_kwargs_from_uri(...).get('url')` already computes the correct value (`parts[:5]`) and still avoids the join, matching the #2599 intent:

```suggestion
        return to_parent_kwargs_from_uri(self.uri).get('url')
```

core/collections/models.py
```diff
@@ -1244,6 +1246,59 @@ def active_mappings(self):
...
+            def key = entry.getKey();
+            def vals = entry.getValue();
+            if (ctx._source[key] == null) { ctx._source[key] = []; }
+            for (v in vals) {
+                if (!ctx._source[key].contains(v)) { ctx._source[key].add(v); }
+            }
+        }
+    """
+
+    def batch_index(self, queryset, document, **kwargs):  # pylint: disable=arguments-differ,unused-argument
```
paynejd, line 1276: **Design gap: this append-only override is also reused on the removal paths, where it can't be correct.**

The Painless script only ever *adds* membership values — it can't drop them. But `index_expansion_concepts` / `index_expansion_mappings` now route through this override on removal too:

- **Exclude references (`add_references`, ~line 1555) — deterministically wrong.** Excluded resources are removed from the M2M, then their ids are re-indexed via `index_resources → index_expansion_concepts → batch_index` (append), which re-adds this collection. There's no corrective full reindex on this path.
- **`delete_references` (~line 1383) — racy.** It fires `batch_index_resources` (full reindex, correct) *and* `self.index_resources(...)` (append, re-adds). Both land on the `indexing` queue; if the append wins, membership is stale.

On `master`, `expansion.batch_index` resolved to the base **full** reindex, which rebuilds from `expansion_set` and correctly drops membership. Net effect after this PR: deleted/excluded concepts still appear when searching or faceting by the collection. #2599 lists partial updates for *adding* collection values but doesn't cover removal.

Suggested direction (not a one-click suggestion — the fix is at call sites outside this hunk):
- In `delete_references`, drop the trailing `self.index_resources(index_concepts, index_mappings)` — the per-ref `batch_index_resources` already does the correct full reindex.
- In `add_references`, reindex *excluded* resources via `batch_index_resources` (full), keeping the fast append only for *included* resources.

Worth adding a remove/exclude → search-visibility test to lock this down.

core/collections/models.py
```diff
@@ -1244,6 +1246,59 @@ def active_mappings(self):
...
+            for rid in batch_ids:
+                yield {
+                    '_op_type': 'update',
+                    '_index': index_name,
+                    '_id': rid,
+                    'script': {
+                        'source': Expansion._APPEND_COLLECTION_FIELDS_SCRIPT,
+                        'params': collection_fields,
+                    },
+                    'upsert': collection_fields,
```
paynejd: **Minor / edge:** `scripted_upsert: True` + `upsert: collection_fields` will create a *skeleton* ES doc containing only these 5 collection fields (no name/id/source/etc.) if the resource isn't already indexed (indexing lag, or `_index=False`). The old full-reindex path always wrote a complete document. Usually the referenced resources are already indexed so this is narrow, but worth a guard or at least a comment noting the assumption.

core/collections/models.py
paynejd: The `/expansions/` handling fixes the **versioned** case cleanly. But there's a remaining edge for **HEAD** collections:

A HEAD collection version's URI is versionless (`/orgs/o/collections/c/`) — `ConceptContainerModel.calculate_uri` only appends the version when `is_versioned and not is_head` (`common/models.py:154`). But the expansion URI carries a literal `HEAD/` segment (`collection_version.uri + 'HEAD/expansions/{mnemonic}/'`, line 1619), so `to_parent_uri(self.uri)` now returns `/orgs/o/collections/c/HEAD/` — with the extra `HEAD/`.

Meanwhile the search filter for HEAD is built *without* a version segment (`/orgs/o/collections/c/`; `common/views.py:507-509` skips the `+= version` when `version == HEAD`) and applied as an exact `term` (`common/views.py:883`). So `.../c/` != indexed `.../c/HEAD/` → **concept/mapping listings for a HEAD collection come back empty.** The old DB-sourced value (`collection_version__uri`) was versionless, so this is a HEAD-only regression, and it's untested (the new `test_batch_index` uses `version='v1'`).

```suggestion
        url = to_parent_uri(self.uri)
        if self.collection_version_name == HEAD:  # HEAD collection-version URIs are versionless
            url = drop_version(url)
        return url
```

(`drop_version('/orgs/o/collections/c/HEAD/')` → `/orgs/o/collections/c/`; versioned URIs are left untouched.) Worth adding a HEAD-collection case asserting the value equals the versionless collection URI. Full disclosure: my original `to_parent_kwargs_from_uri(...).get('url')` suggestion would've had the same gap — the root cause is that the expansion URI inherently contains `HEAD/` while the canonical collection-version URI and the filter omit it.

core/common/mixins.py
```diff
@@ -739,23 +739,31 @@ def owner(self):
...
     def parent_resource(self):
-        return get(self.parent, 'mnemonic')
+        return get(self.parent_url_kwargs, 'repo')
 
     @property
     def parent_url(self):
-        return get(self.parent, 'uri')
+        return to_parent_uri(self.uri)
+
+    @property
```
paynejd: **Minor (efficiency):** `owner_name`, `owner_type`, and `parent_resource` each read `parent_url_kwargs`, so the URI is re-parsed 3× per resource during indexing/serialization. It's derived purely from the (stable) `uri`, so caching collapses that to a single parse — consistent with the `@cached_property` already used in this file (lines 80, 633):

```suggestion
    @cached_property
```

Optional follow-on: `parent_url` could also reuse `self.parent_url_kwargs.get('url')` instead of its own `to_parent_uri(self.uri)` call, making it one parse for all four properties. (`parent_url_kwargs['url']` is computed before the `version=None` mutation, so it's unaffected.)

paynejd (commented):
One more minor efficiency note inline (the correctness item — HEAD `collection_url` — is in the thread above).

PR #889, opened 2026-07-03
OpenConceptLab/ocl_issues#2601 | Qwen model logit score and LRU cache

### LINKED ISSUE #2601: Qwen/Qwen3-Reranker-0.6B memory load and score between 0-1
**Qwen Reranker docs - https://huggingface.co/Qwen/Qwen3-Reranker-0.6B**

### Issues:
1. Scores are not between 0-1.
```
By default, scores are raw logit differences. To get 0-1 probability scores, pass a Sigmoid activation function:

scores = model.predict([(query, doc) for doc in documents], activation_fn=torch.nn.Sigmoid())
```
2. This custom encoder when loaded in memory per request has the potential to crash the service (crashes locally on second request), because every request loads the model again and prev request is not releasing memory cleanly. 

### Solution:
1. use `activation_fn=torch.nn.Sigmoid()` for `Qwen` models
2. custom reranker models should go through a bounded in-process cache instead of being freshly loaded

#### LRU Cache implementation:
- Keeps custom models in a class-level LRU-style cache with a lock.
- Limits custom models to 1 cached model (`settings.RERANKER_CUSTOM_ENCODER_CACHE_SIZE`) by default, so loading Qwen evicts any previously cached custom model before another - large model is loaded.
- Expires idle custom models after 300 seconds (`settings.RERANKER_CUSTOM_ENCODER_CACHE_TTL`) by default.
- Calls eager cleanup (gc.collect(), plus CUDA cache cleanup if applicable) when a custom model is evicted.

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_issues#2601

### DIFF
--- a/core/common/search.py
+++ b/core/common/search.py
@@ -1,6 +1,10 @@
+import gc
+import inspect
 import re
+import threading
 import time
 import urllib
+from collections import OrderedDict
 
 from cid.locals import get_cid
 from django.conf import settings
@@ -8,6 +12,7 @@ from django.db.models import Case, When, IntegerField
 from elasticsearch_dsl import FacetedSearch, Q
 from pydash import compact, get, has, set_
 from sentence_transformers import CrossEncoder
+import torch
 
 from core.common.constants import ES_REQUEST_TIMEOUT
 from core.common.utils import is_url_encoded_string
@@ -337,6 +342,8 @@ class CustomESSearch:
 
 
 class Reranker:
+    """Rerank semantic search hits with model-specific score normalization."""
+
     ENCODERS = [
         # Best and Fastest overall lightweight medical reranker
         # Size: ~110M
@@ -368,6 +375,9 @@ class Reranker:
     ]
     SCORE_KEY = 'search_rerank_score'
     MISSING_SCORE = -1000000.0
+    QWEN_RERANKER_PREFIX = 'Qwen/'
+    CUSTOM_ENCODER_CACHE = OrderedDict()
+    CUSTOM_ENCODER_CACHE_LOCK = threading.Lock()
 
     def __init__(self, model_name=None):
         self.model_name = model_name
@@ -381,6 +391,11 @@ class Reranker:
 
     @property
     def default_model(self):
+        return settings.ENCODER_MODEL_NAME
+
+    @classmethod
+    def _get_default_model_name(cls):
+        """Return the default boot-time reranker model name."""
         return settings.ENCODER_MODEL_NAME
 
     # private
@@ -399,11 +414,29 @@ class Reranker:
                 valid.append((i, d.strip()))
         if not valid:
             return scores_full
-        scores = self.encoder.predict([(txt, d) for _, d in valid])
+        scores = self.encoder.predict([(txt, d) for _, d in valid], **self._get_predict_kwargs())
         for (i, _), s in zip(valid, scores):
             scores_full[i] = float(s)
 
         return scores_full
+
+    def _get_activation_fn(self):
+        """Return the score activation required by the configured reranker model."""
+        model_name = self.model_name or self.default_model
+        if isinstance(model_name, str) and model_name.startswith(self.QWEN_RERANKER_PREFIX):
+            return torch.nn.Sigmoid()
+        return None
+
+    def _get_predict_kwargs(self):
+        """Return compatibility kwargs for CrossEncoder.predict across library versions."""
+        activation_fn = self._get_activation_fn()
+        if activation_fn is None:
+            return {}
+
+        predict_signature = inspect.signature(self.encoder.predict)
+        if 'activation_fct' in predict_signature.parameters:
+            return {'activation_fct': activation_fn}
+        return {'activation_fn': activation_fn}
 
     def _assign_score(self, hits, scores, score_key, order_results):
         score_key = score_key or self.SCORE_KEY
@@ -420,10 +453,31 @@ class Reranker:
     def _order(hits, key_to_order):
         return sorted(hits, key=lambda hit: get(hit, key_to_order), reverse=True)
 
-    def _get_encoder(self, model_name):
-        if model_name and model_name != self.default_model:
-            return self._load_encoder(model_name)
-        return self._load_default_encoder()
+    @classmethod
+    def _get_encoder(cls, model_name):
+        if model_name and model_name != cls._get_default_model_name():
+            return cls._get_custom_encoder(model_name)
+        return cls._load_default_encoder()
+
+    @classmethod
+    def _get_custom_encoder(cls, model_name):
+        """Return a bounded cached custom encoder to avoid repeated large-model loads."""
+        now = time.time()
+        with cls.CUSTOM_ENCODER_CACHE_LOCK:
+            cls._evict_expired_custom_encoders(now)
+            cached_encoder = cls.CUSTOM_ENCODER_CACHE.get(model_name)
+            if cached_encoder:
+                cls.CUSTOM_ENCODER_CACHE.move_to_end(model_name)
+                cached_encoder['expires_at'] = now + cls._get_custom_encoder_cache_ttl()
+                return cached_encoder['encoder']
+
+            cls._evict_custom_encoders_for_capacity()
+            encoder = cls._load_encoder(model_name)
+            cls.CUSTOM_ENCODER_CACHE[model_name] = {
+                'encoder': encoder,
+                'expires_at': now + cls._get_custom_encoder_cache_ttl(),
+            }
+            return encoder
 
     @staticmethod
     def _load_encoder(model_name):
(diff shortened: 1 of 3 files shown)


core/common/search.py
```diff
@@ -399,12 +414,30 @@ def _predict_scores(self, hits, txt, name_key, source_attr, should_convert_sourc
...
+        scores = self.encoder.predict([(txt, d) for _, d in valid], **self._get_predict_kwargs())
         for (i, _), s in zip(valid, scores):
             scores_full[i] = float(s)
 
         return scores_full
 
+    def _get_activation_fn(self):
+        """Return the score activation required by the configured reranker model."""
+        model_name = self.model_name or self.default_model
+        if isinstance(model_name, str) and model_name.startswith(self.QWEN_RERANKER_PREFIX):
```
filiperochalopes: For me this approach is too hardcoded, maybe its interesting to do somethins like:

```
RERANKER_SCORE_PROFILES = {
    "Qwen/": "logit_sigmoid",
    "BAAI/bge-reranker": "logit_sigmoid",
    "cross-encoder/ms-marco-": "raw_rank_score",
    "ncbi/MedCPT-Cross-Encoder": "logit_sigmoid",
}
```

But it also could be too YAGNI, so let's hear @paynejd thoughts

filiperochalopes (commented):
`RERANKER_SCORE_PROFILES` could work as a router for what function to use to normalize the score 0-1. Use `Qwen/` looks too specific. But works

core/common/search.py
snyaggarwal: Valid feedback, added `SIGMOID_MODELS = ['Qwen/']` to accommodate future models
removed `QWEN_RERANKER_PREFIX`

core/common/search.py
```diff
@@ -439,3 +503,41 @@ def _get_source(data, source_attr, should_convert_source_to_dict):
...
+            if cached_encoder:
+                cls._release_encoder(cached_encoder['encoder'])
+
+    @staticmethod
+    def _release_encoder(encoder):
+        """Release model references eagerly to give Python a chance to reclaim memory."""
+        del encoder
+        gc.collect()
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
```
paynejd, line 567: `del encoder` only unbinds the function-local name — at both call sites the model is still referenced by the caller's `cached_encoder['encoder']` when `gc.collect()` runs, so the collector can never reclaim it here (it is freed a moment later by plain refcounting, which needs no GC at all). Net effect: every eviction pays a full-heap GC pass **while holding `CUSTOM_ENCODER_CACHE_LOCK`**, and frees nothing. `torch.cuda.empty_cache()` is also moot for these encoders since `_load_encoder` pins `device="cpu"`.

Dropping the cache entry first and collecting after the last reference is gone keeps the eager-cleanup intent and actually works:

```suggestion
    @classmethod
    def _evict_custom_encoders_for_capacity(cls):
        """Evict least-recently-used custom encoders before loading another large model."""
        while len(cls.CUSTOM_ENCODER_CACHE) >= cls._get_custom_encoder_cache_size():
            cls.CUSTOM_ENCODER_CACHE.popitem(last=False)
            cls._release_memory()

    @classmethod
    def _evict_expired_custom_encoders(cls, now=None):
        """Remove expired custom encoders so idle large models do not stay resident forever."""
        now = now or time.time()
        expired_models = [
            model_name for model_name, cached_encoder in cls.CUSTOM_ENCODER_CACHE.items()
            if cached_encoder['expires_at'] <= now
        ]
        for model_name in expired_models:
            if cls.CUSTOM_ENCODER_CACHE.pop(model_name, None):
                cls._release_memory()

    @staticmethod
    def _release_memory():
        """Reclaim memory after the last cache reference to an evicted encoder has been dropped."""
        gc.collect()
        if torch.cuda.is_available():
            torch.cuda.empty_cache()
```

With this, the two tests that patch `_release_encoder` and assert it received the old encoder would instead patch `_release_memory` and assert it was called once (no args).

core/common/search.py
```diff
@@ -420,10 +463,31 @@ def _assign_score(self, hits, scores, score_key, order_results):
...
+        now = time.time()
+        with cls.CUSTOM_ENCODER_CACHE_LOCK:
+            cls._evict_expired_custom_encoders(now)
+            cached_encoder = cls.CUSTOM_ENCODER_CACHE.get(model_name)
+            if cached_encoder:
+                cls.CUSTOM_ENCODER_CACHE.move_to_end(model_name)
+                cached_encoder['expires_at'] = now + cls._get_custom_encoder_cache_ttl()
+                return cached_encoder['encoder']
+
+            cls._evict_custom_encoders_for_capacity()
```
paynejd: Worth documenting (code comment) here: the size bound caps *cache references*, not resident memory. An in-flight request keeps an evicted encoder alive via `self.encoder`, so under threaded serving (see review summary) a worker can briefly hold old model + new model + the boot-time default simultaneously — the same OOM shape this PR targets, despite `CACHE_SIZE=1`. With sync workers this can't happen; a one-line comment saves the next person the analysis.

core/common/search.py
```diff
@@ -420,10 +463,31 @@ def _assign_score(self, hits, scores, score_key, order_results):
...
+    def _get_custom_encoder(cls, model_name):
+        """Return a bounded cached custom encoder to avoid repeated large-model loads."""
+        now = time.time()
+        with cls.CUSTOM_ENCODER_CACHE_LOCK:
+            cls._evict_expired_custom_encoders(now)
+            cached_encoder = cls.CUSTOM_ENCODER_CACHE.get(model_name)
+            if cached_encoder:
+                cls.CUSTOM_ENCODER_CACHE.move_to_end(model_name)
+                cached_encoder['expires_at'] = now + cls._get_custom_encoder_cache_ttl()
+                return cached_encoder['encoder']
```
paynejd: Returning one shared `CrossEncoder` to concurrent callers: HuggingFace fast tokenizers are not thread-safe (`RuntimeError: Already borrowed` under concurrent `predict`). Inert with today's sync workers (see summary), and the default encoder is already shared this way — but if threaded serving is the plan behind this lock, `predict` will need per-encoder serialization too.

core/common/search.py
```diff
@@ -420,10 +463,31 @@ def _assign_score(self, hits, scores, score_key, order_results):
...
+
+            cls._evict_custom_encoders_for_capacity()
+            encoder = cls._load_encoder(model_name)
```
paynejd: The full model load (multi-second, multi-minute on a cold HF cache) runs inside the global lock, so under threaded serving a cache *hit* for an already-loaded model stalls behind another model's download — hits pay for misses. Only the miss needs exclusivity: double-checked locking (load outside the lock, re-check before insert) or a per-model lock keeps hits fast. Fine to defer while workers are sync (see summary).

core/common/search.py
```diff
@@ -368,6 +375,9 @@ class Reranker:
     ]
     SCORE_KEY = 'search_rerank_score'
     MISSING_SCORE = -1000000.0
+    SIGMOID_MODELS = ['Qwen/']
```
paynejd: This hard-codes a vendor namespace as a proxy for "emits raw logits", and `encoder_model` is a free-form user-supplied string — a re-hosted or finetuned Qwen reranker (e.g. `myorg/qwen3-reranker-ft`) silently misses the sigmoid and `search_normalized_score` becomes `logit*100`. Since this PR already makes the cache env-configurable, consider the same here (e.g. `RERANKER_SIGMOID_MODEL_PREFIXES`, defaulting to `Qwen/`) so a mis-scaled model is a config change rather than a deploy.

core/common/search.py
```diff
@@ -399,12 +414,40 @@ def _predict_scores(self, hits, txt, name_key, source_attr, should_convert_sourc
...
+    def _get_predict_kwargs(self):
+        """Return compatibility kwargs for CrossEncoder.predict across library versions."""
+        activation_fn = self._get_activation_fn()
+        if activation_fn is None:
+            return {}
+
+        predict_signature = inspect.signature(self.encoder.predict)
+        if 'activation_fct' in predict_signature.parameters:
+            return {'activation_fct': activation_fn}
+        return {'activation_fn': activation_fn}
```
paynejd, line 443: The `activation_fct` branch is dead code under the pin: the kwarg was renamed to `activation_fn` in the sentence-transformers 4.0 CrossEncoder rewrite, and requirements.txt pins 5.4.0 (checked the 5.4.0 source — `predict` takes `activation_fn`). If an environment ever drifts below the pin, this shim silently masks the drift instead of surfacing it — and it re-runs `inspect.signature` on every predict call.

```suggestion
    def _get_predict_kwargs(self):
        """Return extra kwargs for CrossEncoder.predict for the configured model."""
        activation_fn = self._get_activation_fn()
        if activation_fn is None:
            return {}
        return {'activation_fn': activation_fn}
```

With this, `import inspect` and `test_reranker_uses_legacy_activation_fct_kwarg_when_needed` go too — suggestions on those lines below.

core/common/search.py
```diff
@@ -1,13 +1,18 @@
+import gc
+import inspect
```
paynejd: Unused once the compat shim in `_get_predict_kwargs` is removed (would fail pylint):

```suggestion
```

core/concepts/tests/tests.py
```diff
@@ -123,6 +124,102 @@ def test_reranker_uses_finite_fallback_score_for_missing_candidate_text(self, ge
...
+            txt='malaria',
+            name_key='name',
+            source_attr='_source',
+            should_convert_source_to_dict=True,
+        )
+
+        self.assertEqual(scores, [0.42])
+        self.assertIsNotNone(encoder.activation_fct)
+        self.assertEqual(encoder.activation_fct.__class__.__name__, 'Sigmoid')
+
```
paynejd: This covers only the pre-4.0 `activation_fct` path, unreachable under the 5.4.0 pin — and once `_get_predict_kwargs` passes `activation_fn` unconditionally, `LegacyEncoder.predict` would raise `TypeError` on the unexpected kwarg, so it has to go together with that change:

```suggestion
```

core/common/search.py
```diff
@@ -383,6 +393,11 @@ def rerank(  # pylint: disable=too-many-arguments
     def default_model(self):
         return settings.ENCODER_MODEL_NAME
```
paynejd: `_get_default_model_name` and this property are now two accessors for the same value, and both are live: `_get_activation_fn` reads `self.default_model` while `_get_encoder` reads `cls._get_default_model_name()`. If default-model resolution ever gains logic, updating one but not the other makes `_get_encoder` classify the real default as custom and load a duplicate copy of the boot model into the cache. Delegate so there is one source of truth:

```suggestion
        return self._get_default_model_name()
```

core/common/search.py
```diff
@@ -439,3 +503,41 @@ def _get_source(data, source_attr, should_convert_source_to_dict):
...
+
+    @classmethod
+    def _get_custom_encoder_cache_size(cls):
+        """Return the max number of custom encoders that may stay loaded per process."""
+        return max(1, int(getattr(settings, 'RERANKER_CUSTOM_ENCODER_CACHE_SIZE', 1)))
+
+    @classmethod
+    def _get_custom_encoder_cache_ttl(cls):
+        """Return the idle TTL for custom encoders in seconds."""
+        return max(1, int(getattr(settings, 'RERANKER_CUSTOM_ENCODER_CACHE_TTL', 60 * 5)))
```
paynejd: These `getattr` fallbacks are dead code — `core/settings.py` now defines both settings unconditionally — and they duplicate the `60 * 5` default in two files, which will silently drift. Direct access matches how this class already reads `settings.ENCODER_MODEL_NAME`, and the tests' `@override_settings` behaves identically:

```suggestion
    @classmethod
    def _get_custom_encoder_cache_size(cls):
        """Return the max number of custom encoders that may stay loaded per process."""
        return max(1, settings.RERANKER_CUSTOM_ENCODER_CACHE_SIZE)

    @classmethod
    def _get_custom_encoder_cache_ttl(cls):
        """Return the idle TTL for custom encoders in seconds."""
        return max(1, settings.RERANKER_CUSTOM_ENCODER_CACHE_TTL)
```

paynejd (commented):
**Review: Qwen sigmoid activation + custom-encoder LRU cache**

What checks out:
- `activation_fn` is the correct kwarg for the pinned sentence-transformers 5.4.0 (verified against the 5.4.0 source), and the sigmoid fix does what ocl_issues#2601 asks.
- The cache genuinely removes the per-request model reload that was crashing the service, and the sliding *idle*-TTL semantics are intentional and documented in the getter.

Shared context for the three threading comments on `_get_custom_encoder`: prod serves with sync gunicorn workers (`startup.sh`, no `--threads`) and celery is prefork, so none of them is triggerable in production today. They matter under dev `runserver` (threaded by default) and become real the day anyone adds `--threads`/gthread — and since the lock exists precisely to anticipate threaded access, the cache should be correct under it. Flagging now while the design is fresh.

PR #897, opened 2026-09-07
Feature/graphql sources 20260906

filiperochalopes:
Solves: https://github.com/OpenConceptLab/ocl_issues/issues/2757

- https://github.com/OpenConceptLab/ocl_issues/issues/2577
- https://github.com/OpenConceptLab/ocl_issues/issues/2356
- https://github.com/OpenConceptLab/oclapi2/pull/816
- https://github.com/OpenConceptLab/oclapi2/pull/838
- https://github.com/OpenConceptLab/oclapi2/pull/877

Why? CIEL Lab needs improvements when loading worklists to load fast using ES-only retrieve for minimal payloads
Also: https://github.com/OpenConceptLab/ocl_issues/issues/2577

### Context

OCL's REST search latency is a bottleneck for several existing flows and blocks new ones from
being viable. Many searches only need a small, fixed payload (a display name, a description, a
few metadata fields) — hydrating a full ORM object and its relations for that is unnecessary
cost. This PR introduces a GraphQL path that serves these minimal payloads straight from
Elasticsearch, without touching the database, while keeping full-object retrieval available
for callers that need it.

### Design principles

- **Minimal payload → no hydration.** A request for a minimal payload should be answerable from
  indexed filters alone, without a database round-trip. Whether to hydrate, and how much, is a
  per-request decision driven by which fields the caller (source or concept payload) actually
  selects — not a fixed retrieval mode for the whole query.
- **Share search classes with REST.** Reuse REST's existing Elasticsearch query/filter
  building blocks instead of duplicating them for GraphQL.
- **Share authorization classes with REST.** Visibility and permission rules stay centralized
  in one place and apply identically to both REST and GraphQL, wherever that's possible.
- **Retrieval path may still differ.** Because the rules that apply depend on the dynamic
  payload requested, GraphQL's retrieval path (index-only vs. hydrated) can legitimately differ
  from REST's, even while reusing REST's search and authorization building blocks.

Documentation is free and detailed

image

image

image

### ⚠️ Breaking change — reindex before deploy

Four fields are new on the search documents: `ConceptDocument.is_active`, `.is_head`,
~`.display_name`~, and `SourceDocument.is_active`. They're additive to the mapping, but existing
documents don't have them populated, and Elasticsearch can't backfill them on its own.

**If you deploy without reindexing first, `concepts` searches silently return zero results.**
The index projection filters on `is_active`/`is_head`; old documents don't match, and a
zero-hit Elasticsearch response is not an error — it's `{"totalCount": 0, "results": []}`,
indistinguishable from a legitimate empty search. (`source` degrades safely instead: missing
`is_active` there falls back to the database automatically.)

Reindex before routing traffic:

```sh
docker exec <api-container> python manage.py search_index --populate --models sources.Source concepts.Concept -f --parallel
```

Treat this as a pre-deploy gate, not cleanup — there's no error to notice afterward.

### New GraphQL API surface

```graphql
query Dictionary($org: String!, $source: String!, $version: String) {
  source(org: $org, source: $source, version: $version) {
    name description canonicalUrl uri
    classes datatypes mapTypes
    externalSources { name url }
    summary { activeConcepts mappings }
  }
}

query FindConcepts($org: String, $source: String, $query: String!, $page: Int, $limit: Int) {
  concepts(org: $org, source: $source, query: $query, page: $page, limit: $limit) {
    totalCount hasNextPage versionResolved
    results { conceptId display description conceptClass datatype { name } }
  }
}
```

- `conceptIds` does exact, case-sensitive mnemonic matching, dedupes, preserves order, and
  takes precedence over `query`. Omit both `org` and `source` for global search.
- `page`/`limit` go together; result window is 10,000 (`totalCount` still reflects the true total).
- Omitted `version` resolves to `HEAD`, falling back to the latest release only if `HEAD` is
  absent; an explicit missing version does not fall back.
- Same auth as REST (OCL token, OIDC bearer, session); anonymous callers see public data only.

### Permissions

Reuses the existing REST visibility rule directly. Concept visibility relies on the indexed
`public_can_view` flag, already kept in sync from the parent repository by
`core/sources/signals.py` whenever the repository's access changes.

### Not indexed, by design

- **`description`** — locale-resolved per concept; selecting it routes the whole request
  through the ORM.
- **`uri`** — rebuilt at query time via Django's `reverse()` from indexed `owner`,
  `owner_type`, `mnemonic`, `version`, reproducing the ORM's percent-encoding exactly (verified
  against all sources in the database, including versions with reserved characters).

### Compatibility

No database migrations, no new environment variables. Existing REST search relevance and
behavior are unchanged.

### Testing

68 GraphQL tests + 249 concepts/sources regression tests passing. Integration suite runs
against real Elasticsearch with temporary indexes, covering zero-SQL projections, owner
isolation, HEAD/release selection, inactive/retired filtering, private-repository visibility,
the rebuilt source URI, and the database fallback for `description`. Pylint clean (10.00/10).


filiperochalopes:
Still draft, but updated, please consider check @snyaggarwal

core/graphql/indexed.py
```diff
@@ -0,0 +1,119 @@
...
+
+logger = logging.getLogger(__name__)
+CONCEPT_INDEX_FIELDS = {
+    '__typename': (),
+    'datatype.__typename': ('datatype',),
+    'id': (),  # Elasticsearch's metadata ID is the OCL database primary key.
+    'conceptId': ('id',),
+    'externalId': ('external_id',),
+    'display': ('display_name',),
+    'description': ('preferred_description',),
```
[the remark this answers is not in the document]
filiperochalopes: Don't think we have use case for it. Seems harmless, but I will create a ticket for it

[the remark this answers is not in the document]
filiperochalopes: https://github.com/OpenConceptLab/ocl_issues/issues/2760

core/concepts/documents.py
```diff
@@ -12,6 +12,11 @@ class Index:
         name = 'concepts'
         settings = {'number_of_shards': 1, 'number_of_replicas': 0}
 
+    # Preserve ORM semantics for direct GraphQL projections without changing REST search fields.
+    is_active = fields.BooleanField(attr='is_active')
+    is_head = fields.BooleanField()
+    display_name = fields.TextField(attr='display_name')
```
snyaggarwal: `display_name` is already indexed in `name` and `_name` with some variations (hangover from ES7)
This adds a next layer on existing confusion on which field to use for name -- `name` or `_name` or `display_name`

1. Can we work with one of the present ones `name` or `_name`?
2. If not, then populate it in `prepare` method where `preferred_locale` is already computed (avoid duplicate compute) and then this becomes `data['display_name'] = get(preferred_locale, 'name')`

core/concepts/documents.py
```diff
@@ -267,3 +272,8 @@ def get_mapped_codes(instance):
                 else:
                     other_mapped_codes.append(to_concept_code)
         return same_as_mapped_codes, other_mapped_codes, verbose_info
+
+    @staticmethod
+    def prepare_is_head(instance):
```
snyaggarwal: there is an existing property `is_head`

snyaggarwal (commented):
@filiperochalopes partial quick review

core/concepts/documents.py
filiperochalopes: The problem using name and _name I see was it is normalized and not represents the display itself

core/concepts/documents.py
filiperochalopes: In the database, right? Not indexed to be quick filtered. I will double check but I was sure of that

snyaggarwal: its a python property in code -- that you can directly reference here in ConceptDocument

filiperochalopes: Yeah. I see. Done

core/concepts/documents.py
filiperochalopes: I got an idea. Simple but useful. I won't use on of but both to undo normalization process and recreate the real display. https://github.com/OpenConceptLab/oclapi2/pull/897/changes#diff-bce917600e5a377a8e5dad8adf538e6726153681f754cf3f2297538cd01b5eafR127

PR #898, opened 2026-09-11
OpenConceptLab/ocl_issues#2760 | reindex concepts on source locale changes

### LINKED ISSUE #2760: Reindex concepts when source locale configuration changes
## User story
As a terminology maintainer, I want concept search projections to stay accurate after I change a source's locale configuration, so that GraphQL and other index-backed consumers do not serve outdated display names.

## Use case
A curator changes a source's `default_locale` or `supported_locales`. Existing concepts under that source keep serving their previously indexed `display_name` in GraphQL, even though the ORM now resolves a different preferred name for the same concept.

## Requirements
- [ ] Reindex a source's concepts when the source's locale configuration changes.
  - [ ] Detect changes to `default_locale`.
  - [ ] Detect changes to `supported_locales`.
- [ ] Analysis: decide where this reindex trigger belongs.
  - [ ] Option A: extend the existing reindex check in `persist_changes` (`core/common/models.py`), which currently reindexes concepts only on `released` or `has_semantic_match_algorithm` changes.
  - [ ] Option B: extend the propagation signal in `core/sources/signals.py`, which currently only propagates `is_active` and `public_access` to concepts.
- [ ] Do not change reindex behavior for any other field.

## Acceptance criteria
- [ ] Given a source with concepts already indexed, when the source's `default_locale` changes, then those concepts are reindexed and their indexed `display_name` reflects the new locale resolution.
- [ ] Given a source with concepts already indexed, when the source's `supported_locales` changes, then those concepts are reindexed and their indexed `display_name` reflects the new locale resolution.
- [ ] Given no locale configuration change, when a source is saved, then no unrelated concept reindex is triggered.

## More details (collapsible)
<details>
<summary>Context and origin</summary>

Found during code review ([chatgpt-codex-connector](https://github.com/OpenConceptLab/oclapi2/pull/897#discussion_r3952531284) bot) on the GraphQL permission-aware source and concept projections PR, on `core/graphql/indexed.py`.

Original finding: "When a source's default_locale or supported_locales changes, these projections continue serving the previously indexed display_name and preferred_description, even though both values are derived from the parent source's locale configuration. Source persistence only reindexes concepts for release or match-algorithm changes, and the new signal propagation handles only activity and visibility, so a lean query requesting display or description remains stale indefinitely while the ORM path returns the new locale selection."

Note: the original finding also mentioned `preferred_description`. That field was removed from the index in a later revision of the same PR (concept descriptions are locale-resolved per concept and are now served through the database fallback instead of being indexed), so this issue covers `display_name` only.

Confirmed in code:
- `Concept.display_name` (core/concepts/models.py) resolves via `preferred_locale`, which reads `self.parent.default_locale` and `self.parent.supported_locales`.
- `persist_changes` (core/common/models.py) only sets `should_reindex_resources` on `released` change and `should_reindex_concepts_only` on `has_semantic_match_algorithm` change. Locale fields are not checked.
- `core/sources/signals.py` `propagate_parent_attributes` only propagates `is_active` and `public_access` to concepts, not locale configuration.

Severity: Warning (non-blocking). Stale display names are a data-freshness issue, not a security or correctness-of-access issue, and only affect the index-only GraphQL projection path. The ORM-hydrated path already resolves the correct value on every request.

Suggested priority: P2, matching the original review comment.
</summary>
</details>

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_issues#2760


filiperochalopes (approved):
The term `exclude_locale` looks confuse to me and I would add an inline comment. I was think more about `default_locale` and not `supported_locales` but you did well covering both.

Besides that it looks good!

snyaggarwal:
> The term `exclude_locale` looks confuse to me and I would add an inline comment. I was think more about `default_locale` and not `supported_locales` but you did well covering both.
> 
> Besides that it looks good!

There are tests to explain "exclude_locale" -- just indexing optimisation.

filiperochalopes:
Yes. I see why. And I didn't figure out another way to do the same. Because of it, I approved.

PR #900, opened 2026-09-21
Capabilities, Permissions and Groups + Mapper preview

snyaggarwal:
## Linked Issue

Refs

https://github.com/OpenConceptLab/ocl_issues/issues/2762
https://github.com/OpenConceptLab/ocl_issues/issues/2780
https://github.com/OpenConceptLab/ocl_issues/issues/2782
https://github.com/OpenConceptLab/ocl_issues/issues/2781


paynejd (commented):
Reviewed against the four linked tickets (#2762, #2780, #2781, #2782). Not approving yet — there are four things I'd want resolved before this can go out in front of the preview gate.

The architecture is sound, and the docstrings explaining *why* (the `MapProjectCapacityExceeded` numeric-detail rationale, the no-implicit-unlimited fallback, `created_by`-not-owner counting) are genuinely better than the repo average. The problems cluster in one seam: **the gate and the meter disagree about what "used" means.**

**Blockers**

1. `POST` a project, delete it, `POST` again → **HTTP 500 with the project already persisted**. The gate reads a live count, the meter is monotonic. (`core/map_projects/views.py`)
2. **Any authenticated user can refund their own quota** by posting negative `units` to `/capabilities/consume/`. (`core/capabilities/views.py`)
3. **The kill switch is inverted** — #2782 says "set caps to 0", but `0` means unlimited in the code. Following the runbook during an incident would uncap everyone.
4. **The migration keys the `preview` group on a hardcoded `pk=16`**, which can hijack an unrelated group's permissions or block startup under `set -e`.

**Also needs a decision before the gate opens**

- No backfill for the 14 existing `mapper-approved` users (#2781) — they lose Mapper access on deploy. The `set_groups()` hazard flagged in #2780 isn't addressed either.
- `mapper.rows_per_project` is metered per-user-for-life though #2762 scopes it per-project; a user with 2+ projects gets permanently locked out of the second one.
- `$match` consumes `units=len(rows)`, but #2762/#2782 specify **row-algorithm pairs** — a 4-algorithm run meters 4x under budget.
- `$match` consumes before `filter_queryset()`, so an ES failure burns unrefundable units — contrary to the consume-after-success rule this PR's own docstring states.
- Org-owned map projects become read-only and undeletable, not just uncreatable.
- Staff (non-superuser) are exempt from caps but not from permissions, so they 403.
- `capabilities.yaml` is reloaded by `pre_startup.sh` on every restart, reverting any group-level cap tuning.

**Scope questions**

- **Demo-project exemption is absent** — no `demo` handling anywhere in the diff, but #2782 requires demo projects exempt from all four caps and locked (no delete, no row change, no config change). Follow-up slice, or an oversight?
- #2782 asks for tests covering **both standalone and Keycloak-replicated modes**; the new tests look standalone-only. Could you confirm?

Details inline. Happy to pair on the gate-vs-meter design call — items 1 and 3 are really the same question about whether the ledger or the live count is authoritative, and I don't think it should be settled in review comments alone.

core/capabilities/views.py
```diff
@@ -0,0 +1,175 @@
...
+    caller authenticated with, so consumption is always scoped to `request.user`;
+    no service can consume against another user's ledger.
+    """
+
+    def post(self, request):
+        capability_name = request.data.get('capability')
+        capability_id, error_response = self.get_object_id()
+        if error_response:
+            return error_response
+        units = int(request.data.get('units', 1))
```
paynejd: **Blocker (security): any authenticated user can refund their own quota.**

`units` is unvalidated, so `{"capability": "mapper.match_operations", "units": -100}` passes the `counter.used + units > limit` check in `check_and_consume_capability` and lands as `F('used') + (-100)` — a self-service ledger reset that bypasses the staff-only `CapabilityRefundView` entirely. A non-numeric `units` raises an uncaught `ValueError` → 500.

```suggestion
        try:
            units = int(request.data.get('units', 1))
        except (TypeError, ValueError):
            return Response({'detail': '"units" must be an integer.'}, status=status.HTTP_400_BAD_REQUEST)
        if units < 1:
            return Response({'detail': '"units" must be a positive integer.'}, status=status.HTTP_400_BAD_REQUEST)
```

core/capabilities/views.py
```diff
@@ -0,0 +1,175 @@
...
+        from core.users.models import UserProfile
+        user = UserProfile.objects.filter(username=username).first()
+        if not user:
+            return Response(status=status.HTTP_404_NOT_FOUND)
+
+        capability_name = request.data.get('capability')
+        capability_id, error_response = self.get_object_id()
+        if error_response:
+            return error_response
+        units = int(request.data.get('units', 1))
```
paynejd: Same unguarded `int()` here — non-numeric `units` is a 500. This view is staff-only so the negative-units path isn't a privilege issue, but a negative refund silently *charges* the user, which is the opposite of what the endpoint advertises.

```suggestion
        try:
            units = int(request.data.get('units', 1))
        except (TypeError, ValueError):
            return Response({'detail': '"units" must be an integer.'}, status=status.HTTP_400_BAD_REQUEST)
        if units < 1:
            return Response({'detail': '"units" must be a positive integer.'}, status=status.HTTP_400_BAD_REQUEST)
```

core/users/migrations/0036_seed_mapper_permissions_and_default_group.py
```diff
@@ -0,0 +1,69 @@
...
+    permissions_by_codename = {}
+    for codename, name in PERMISSIONS:
+        permission, _created = permission_model.objects.get_or_create(
+            codename=codename, content_type=content_type, defaults={'name': name}
+        )
+        permissions_by_codename[codename] = permission
+
+    preview_group, _created = group_model.objects.get_or_create(
+        pk=PREVIEW_GROUP_PK, defaults={'name': PREVIEW_GROUP_NAME}
+    )
```
paynejd: **Blocker: keying the group on a hardcoded pk is unsafe on prod.**

The rationale in the comment above is legitimate — the pk pin does need to dodge `auth_groups.yaml`'s 1–15 on a fresh test database. But `get_or_create(pk=16, defaults={'name': ...})` keys on the **pk**, not the name, which gives two bad outcomes on an environment where pk 16 isn't what we expect:

- pk 16 is some other group → it's matched as `preview` and `permissions.set()` **wipes and replaces that group's permissions**.
- a `preview` group already exists at a different pk → `create` hits the unique-name `IntegrityError`, and since `pre_startup.sh` runs under `set -e` that blocks container startup.

Keying on `name` keeps the migration idempotent and still pins the pk in the fresh-test-DB case:

```suggestion
    preview_group = group_model.objects.filter(name=PREVIEW_GROUP_NAME).first()
    if not preview_group:
        # Pin the pk only when it is actually free, so the fixture-collision guard
        # above still holds on a fresh test database without risking a pk clash on
        # an environment that already has something at PREVIEW_GROUP_PK.
        defaults = {} if group_model.objects.filter(pk=PREVIEW_GROUP_PK).exists() else {'pk': PREVIEW_GROUP_PK}
        preview_group = group_model.objects.create(name=PREVIEW_GROUP_NAME, **defaults)
```

Worth confirming against prod's `auth_group` before merge either way.

core/common/permissions.py
```diff
@@ -97,10 +102,112 @@ def has_object_permission(self, request, view, obj):
...
+    mapper.projects: how many map projects a user may have. Only meaningful at
+    creation time - unlike the other three Mapper gates this isn't a plain
+    permission check but a numeric usage comparison, so it lives here instead
+    of as a view-method-level cap check.
     """
     def has_permission(self, request, view):
         user = request.user
-        return user.is_authenticated and user.has_auth_group(MAPPER_AI_ASSISTANT_GROUP)
+        limit = user.get_capability_limit(MAPPER_PROJECTS_CAPABILITY_ID)
+        if limit == 0:  # explicit grant only - None (unconfigured) is blocked, not uncapped
```
paynejd: **Blocker (operational): the documented kill switch is inverted.**

ocl_issues#2782 states the kill switch is data — *"set the default role's caps to 0"*. In this implementation `0` means **unlimited** (`get_capability_limit` docstring, and the `if limit == 0: return True` here), while `None` means blocked. An operator following the ticket during an incident would hand every preview user unlimited access.

The convention itself is defensible and well-documented, so I'd rather fix the ticket than the code — the actual kill is deleting the `GroupCapability` row so the limit resolves to `None`. But that needs to be true in exactly one place. Could you either:

1. update #2782's AC to describe the real mechanism, and add a line here naming it (`# kill switch = remove the GroupCapability row, which resolves to None = blocked`), or
2. introduce a distinct sentinel for unlimited (e.g. `-1`) so `0` means what everyone will assume it means?

My preference is (1) plus the comment — but it shouldn't merge with the runbook and the code disagreeing.

core/common/permissions.py
```diff
@@ -97,10 +102,112 @@ def has_object_permission(self, request, view, obj):
...
+    For map-project creation: the parent (org or user) the project would be
+    created under isn't an object DRF's generic view machinery resolves via
+    get_object() - it's resolved onto the view by ConceptDictionaryCreateMixin.
+    set_parent_resource(), which normally only runs from post()/create(), after
+    permissions are checked. Resolve it here too so the ownership check (same
+    rule as HasOwnership.has_object_permission) runs as a request-level
+    permission instead of view-method code.
+    """
+    def has_permission(self, request, view):
+        view.set_parent_resource()
```
paynejd, line 188: **Bug: this makes `POST /user/map-projects/` always 403.**

`set_parent_resource()` (`core/common/mixins.py:432`) falls back to `self.request.user.username` only when `self.user_is_self` is true. But `user_is_self` is assigned in `BaseAPIView.initialize()`, which runs *after* `super().initial()` — and `super().initial()` is what calls `check_permissions()`:

```python
# core/common/views.py:120
def initial(self, request, *args, **kwargs):
    super().initial(request, *args, **kwargs)   # <- check_permissions() runs here
    self.initialize(request, request.path_info, **kwargs)  # <- user_is_self set here
```

So at permission-check time `view.user_is_self` is still the class default `False`. On the `/user/map-projects/` route (`core/users/user_urls.py:72`, which supplies `extra_kwargs = {'user_is_self': True}`) neither `org` nor `user` is in `self.kwargs`, `parent_resource` resolves to `None`, and this returns `False`.

Scope check: oclmap posts to `/users/<username>/map-projects/` (`MapProjects.jsx:55`, `MapProject.jsx:1505`), which passes `user` explicitly — so the Mapper UI isn't broken by this. It's the `/user/` alias that dies, and it's a published route.

Simplest fix is to read the kwarg directly rather than the not-yet-set attribute, e.g. have `set_parent_resource()` fall back to `self.kwargs.get('user_is_self')`. A regression test hitting `POST /user/map-projects/` would be worth adding — the current tests appear to go through the explicit-username route only.

core/common/permissions.py
```diff
@@ -97,10 +102,112 @@ def has_object_permission(self, request, view, obj):
...
+    don't instantiate directly. `message` is set dynamically so DRF's permission-denied
+    response carries a structured `error_code`, not just a generic string `detail`.
+    """
+    capability_name = None
+    error_code = 'mapper_access_denied'
+    denied_message = 'You do not have access to the Mapper.'
+
+    def has_permission(self, request, view):
+        user = request.user
+        if user and user.is_authenticated and user.has_perm(self.capability_name):
```
paynejd: **Inconsistency: staff are exempt from caps but not from permissions.**

`has_perm()` short-circuits for superusers only, whereas `get_capability_limit()` (`core/users/models.py:263`) exempts `is_superuser or is_staff`. A staff-but-not-superuser account therefore gets uncapped quota and a 403 at the gate — it can't reach the endpoint it has unlimited budget for.

Pick one rule and apply it in both places. Given the `get_capability_limit` docstring calls staff *"the deliberate exception"*, exempting staff here too looks like the intent.

core/map_projects/views.py
```diff
@@ -45,13 +71,15 @@ def get(self, request, *args, **kwargs):
...
-            return Response(status=status.HTTP_403_FORBIDDEN)
+
         serializer = self.get_serializer(data=MapProject.format_request_data(request.data, self.parent_resource))
         if serializer.is_valid():
             instance = serializer.save(force_insert=True)
             if serializer.is_valid():
+                # The permission class checks capacity before creation; record actual usage only after save succeeds.
+                request.user.check_and_consume_capability(
+                    MAPPER_PROJECTS_CAPABILITY_ID, action='create_map_project', map_project=instance
+                )
```
paynejd: **Blocker: delete-then-recreate returns a 500 with the project already persisted.**

The gate and the meter disagree about what "used" means:

- `HasMapProjectCapacity` (`core/common/permissions.py:210`) reads `user.map_projects_used`, a **live count** of `is_active=True` projects.
- This call consumes into `UsageCounter.used`, which is **monotonic** and never decremented on delete.

With `mapper.projects = 1`:

1. Create project A — gate passes (live `0 < 1`), counter `0 → 1`.
2. Delete A — live count back to 0, counter stays 1.
3. Create project B — gate passes (live `0 < 1`), `serializer.save(force_insert=True)` **commits B**, then this call raises `CapabilityExceeded` (`1 + 1 > 1`).

`CapabilityExceeded` subclasses plain `Exception`, not `APIException`, so it isn't caught here and isn't handled by DRF — the user gets a **500**, and project B is already in the database because `save()` committed outside the `transaction.atomic()` block inside `check_and_consume_capability`.

ocl_issues#2782 says *"Deleting a project frees the slot and refunds nothing"*, so the live-count gate is the spec and the monotonic consume is the defect. Options: drop the consume here and keep `map_projects_used` as the single source of truth (the `UsageEvent` row is still worth writing for attribution), or make `mapper.projects` fully monotonic on both sides — but not one of each.

Separately: `CapabilityExceeded` escaping as a 500 is a general hazard, not specific to this call site. Either make it an `APIException` like its sibling `MapProjectCapacityExceeded`, or add a DRF exception handler for it.

core/map_projects/views.py
```diff
@@ -223,5 +277,34 @@ def get_object(self, queryset=None):  # pylint: disable=arguments-differ
...
+        serializer.is_valid(raise_exception=True)
+
+        completed_rows = serializer.validated_data.get('completed_rows')
+        completed_rows_delta = max((completed_rows or 0) - instance.completed_rows, 0)
+        try:
+            with transaction.atomic():
+                self.perform_update(serializer)
+                if completed_rows_delta:
+                    request.user.check_and_consume_capability(
+                        MAPPER_ROWS_PER_PROJECT_CAPABILITY_ID, units=completed_rows_delta,
```
paynejd: **`mapper.rows_per_project` is metered per-user for life, but it's specified as project-scoped.**

ocl_issues#2762 defines the scope table explicitly: `mapper.projects` → user, **`mapper.rows_per_project` → project**. The check at line ~223 does honour that (it's evaluated against the run/project), but consumption here goes into the per-user, never-reset `UsageCounter`.

For a grandfathered user (ocl_issues#2781: 9 of 13 creators have more than one project) the counter accumulates across every project they own, so once lifetime rows exceed the limit they're permanently 403'd on progress PATCHes in **every** project — including brand new ones with zero rows.

If the ledger is going to stay user-keyed, this cap needs a project-scoped counter (the `UsageEvent` already carries `map_project`, so the data model supports it).

core/map_projects/views.py
```diff
@@ -20,6 +25,27 @@ class MapProjectBaseView(BaseAPIView):
...
+                HasMapProjectParentOwnership, CanCreateOrgMapProjects, HasMapProjectCapacity,
+                CanUseCustomMapperAlgorithms,
+            )
+        elif method in ('PUT', 'PATCH'):
+            permission_classes = (CanEditConceptDictionary, CanCreateOrgMapProjects, CanUseCustomMapperAlgorithms)
+        elif method == 'DELETE':
+            # A user who couldn't create an org-owned project shouldn't be able to
+            # delete one either; CanEditConceptDictionary still gates on the
+            # project itself via get_object()/check_object_permissions.
+            permission_classes = (CanEditConceptDictionary, CanCreateOrgMapProjects)
```
paynejd: **This makes existing org-owned projects read-only and undeletable.**

`CanCreateOrgMapProjects` requires `mapper_org_projects`, which no group holds — not even `preview` (the 0036 migration grants only `mapper_use` and `mapper_ai_assistant`). Applied to PUT/PATCH (line 39) and DELETE here, that means every org-owned map project becomes immutable for non-superusers.

ocl_issues#2782 scopes this as *"org-owned projects unavailable in the preview"*, which I read as blocking **creation**. Prod has 1 org-owned project today (#2781); bricking it probably isn't intended. Suggest gating creation only, and letting existing org projects stay editable/deletable by their owners.

Minor, same area: the permission is spelled `mapper_org_projects` here vs `mapper.org_projects` in #2780's AC. Worth aligning the ticket or the codename so the two are greppable together.

core/concepts/views.py
```diff
@@ -1100,16 +1141,27 @@ def get_repo_params(is_semantic, target_repo_params, target_repo_url):
...
-        if user.is_mapper_waitlisted or not user.is_mapper_approved:
-            return Response(
-                {'detail': 'You are currently in waitlist for $match operation.'},
-                status=status.HTTP_403_FORBIDDEN
-            )
+        rows = request.data.get('rows')
+        if isinstance(rows, list) and rows:
+            algorithm_id, map_project = get_match_operations_attribution(request)
+            try:
+                request.user.check_and_consume_capability(
```
paynejd, line 1157: **Consumes quota before the work succeeds — the PR's own rule says not to.**

`check_and_consume_capability`'s docstring states a caller *"should consume a capability only once the action it gates has actually succeeded ... which needs no refund path at all"*. Here consumption happens before `self.filter_queryset()` at line 1165, so an ES timeout or any downstream failure burns the user's units with no user-reachable refund (`CapabilityRefundView` is `IsAdminUser`). For a 100-operation lifetime allowance with no reset, a bad ES day is unrecoverable without staff intervention.

**Also the meter unit looks wrong.** This consumes `units=len(rows)` against a single `algorithm_id`, but #2762 and #2782 both specify **row-algorithm pairs** — #2782 spells it out: *"one AutoMatch run may fire ... several requests per row (one per configured algorithm) ... So 25 rows can be 25 match operations or 100+."* As written, a 4-algorithm run over 25 rows meters 25, not 100, so the 100-operation cap is effectively 4x looser than designed for multi-algorithm users. Is that a deliberate simplification for the preview? If so it's worth saying so in #2762 and adjusting the starting value.

core/concepts/views.py
```diff
@@ -880,12 +885,48 @@ def post(request):
...
+
+    algorithm_id = metadata.get('algorithm_id')
+    if algorithm_id is not None:
+        algorithm_id = str(algorithm_id)[:100]  # UsageEvent.algorithm is a CharField(max_length=100)
+
+    map_project = None
+    map_project_id = metadata.get('map_project_id')
+    if isinstance(map_project_id, (str, int)) and str(map_project_id).isdigit():
+        from core.map_projects.models import MapProject
+        map_project = MapProject.objects.filter(id=int(map_project_id)).first()
```
paynejd: **No ownership check on a client-supplied project id.**

`map_project_id` comes straight from the `X-OCL-Event-Metadata` request header and is resolved to any `MapProject` by id, so a caller can attribute their `UsageEvent` rows to a project they don't own. Usage stays correctly charged to `request.user`, so this isn't a quota bypass — but it does corrupt the per-project attribution that ocl_online#105 is building billing on.

Cheap fix: scope the lookup to projects the user can see, e.g. filter on `created_by=request.user` (or the same ownership rule the map-project views use) and fall back to `None` rather than trusting the id.

core/users/constants.py
```diff
@@ -10,8 +10,8 @@
 HAPI_FHIR_SERVERS_GROUP = 'hapi_fhir_servers'
 OPERATIONS_PANEL_GROUP = 'operations_panel'
 MAPPER_AI_ASSISTANT_GROUP = 'mapper_ai_assistant'
-MAPPER_WAITLIST_GROUP = 'mapper-waitlist'
-MAPPER_APPROVED_GROUP = 'mapper-approved'
+MAPPER_WAITLIST_GROUP = 'mapper-waitlist'  # obsolete no effect
+MAPPER_APPROVED_GROUP = 'mapper-approved'  # obsolete no effect
```
paynejd, line 14: **No backfill: every currently approved Mapper user loses access on deploy.**

This PR references ocl_issues#2781, but I don't see a data migration moving `mapper-approved` members into `preview` (or granting them the unlimited caps #2781 asks for). `mapper_use` is only granted via the `preview` group created in migration 0036, and nothing adds existing users to it — so on deploy the 14 `mapper-approved` accounts, which cover all 13 project creators, get 403 on `$match`/`$rerank` and project creation.

#2780 also requires *"a re-runnable backfill for existing accounts"* for the default preview role, and #2781 additionally wants the grant verified by an **SSO login round-trip** — `core/common/backends.py:69,80` calls `user.set_groups(claims.get('groups', []))` on every OIDC login, which replaces the entire Django group set and would silently undo a Django-side backfill. That hazard is called out in #2780 and isn't addressed in this diff.

Is the backfill intended as a separate PR? If so, worth noting in the description so it can't ship ahead of this one.

core/fixtures/capabilities.yaml
```diff
@@ -0,0 +1,44 @@
...
+    description: Number of concept-match operations (rows x algorithms)
+- model: "capabilities.capability"
+  pk: 4
+  fields:
+    name: ai_assistant.calls
+    description: Number of Mapper AI Assistant invocations
+- model: "capabilities.groupcapability"
+  pk: 1
+  fields:
+    group: 16  # preview - core/fixtures/auth_groups.yaml
```
paynejd: **These limits revert on every container restart.**

`pre_startup.sh` runs `loaddata core/fixtures/*` on every start, and these rows are pk-pinned, so any operational tuning of a group's caps via the admin gets silently overwritten the next time the container cycles. That conflicts with #2762's *"Admins can raise any cap per user or per role"* — and it also means the kill switch (whatever form it takes) can't be applied at group level durably.

Per-user `UserCapabilityOverride` rows survive, so there is a workaround, but it doesn't scale to "raise the cap for the preview role". Options: drop `GroupCapability` rows from the fixture and seed them in the migration instead, or narrow what `pre_startup.sh` loads.

Also note this file hardcodes `group: 16` in four places, which inherits whatever we decide about the pk pinning in migration 0036.

core/capabilities/views.py
```diff
@@ -0,0 +1,175 @@
...
+
+        try:
+            request.user.check_and_consume_capability(
+                capability_id, units=units,
+                action=request.data.get('action', ''), algorithm=request.data.get('algorithm'),
+            )
```
paynejd, line 96: **`map_project` and `run` are dropped here, so cross-service consumption lands in `UsageEvent` with both FKs null.**

`check_and_consume_capability` accepts `map_project=` and `run=`, and `UsageEvent` has real FK columns for them — but this view forwards only `action` and `algorithm`, so nothing a cross-service caller sends for those can reach the ledger.

That matters because ocl-ai-assistant is the caller this view exists for, and #2762's AC is explicit: *"Per-action consumption records (action, units, algorithm, project, run) so configuration-based weighting can be added later without a rewrite."* Without project/run, `$invoke` consumption can't be attributed to a project or an AutoMatch run, which is most of what #105/#114 want the ledger for. The in-process `$match` path gets this right; the HTTP path silently doesn't.

The gap is symmetric — ocl-ai-assistant#195 passes only `action='invoke'` today, even though `core/common/attribution.py` already parses `automatch_run_id` and `map_project_id` off every request. I've raised that half there. Both need the same field names agreed, so worth settling here first:

```suggestion
        try:
            request.user.check_and_consume_capability(
                capability_id, units=units,
                action=request.data.get('action', ''), algorithm=request.data.get('algorithm'),
                map_project_id=request.data.get('map_project'), run_id=request.data.get('run'),
            )
```

(`check_and_consume_capability` would take `map_project_id`/`run_id` alongside the existing object kwargs, so a caller that only has ids doesn't need to fetch the rows. Both should be validated as belonging to `request.user` before they're written, or a caller could attribute its usage to someone else's project.)

paynejd (commented):
Follow-up from reviewing the ai-assistant half — OpenConceptLab/ocl-ai-assistant#195 (review posted there now). One cross-PR gap that can only be closed on this side, plus the merge-sequencing consequence.

**Sequencing:** #195 sends `includeCapabilities=true` and reads `permissions`/`capabilities`, and both its gates fail closed when those fields are absent. So #195 cannot ship before this PR — and this PR cannot ship before the #2781 backfill, because `users.mapper_ai_assistant` is granted only to the new `preview` group and nothing here puts anyone into it. Concretely, on deploy day `$invoke` 403s for the 8 `mapper_ai_assistant` members, the 7 `core_user` members and all non-superuser staff; only superusers keep access, since Django grants superusers every permission. That's the same missing-backfill item from my earlier review, but it now has a second repo depending on it. If the backfill slips, the interim unblock is granting `users.mapper_ai_assistant` to the existing `mapper_ai_assistant` group (pk 5).

Detail on the attribution gap inline.

core/capabilities/views.py
paynejd: **B2.** `units` is unbounded, and negative values decrement the ledger (`update(used=F('used') + units)`), so any authenticated user can unlimit themselves with one request. A non-numeric value is also an unhandled `ValueError` -> 500.

```suggestion
        try:
            units = int(request.data.get('units', 1))
        except (TypeError, ValueError):
            return Response({'detail': '"units" must be an integer.'}, status=status.HTTP_400_BAD_REQUEST)
        if units < 1:
            return Response({'detail': '"units" must be a positive integer.'}, status=status.HTTP_400_BAD_REQUEST)
```

`check_and_consume_capability` should also reject non-positive `units` independently, since OpenConceptLab/ocl_issues#2762 makes it the shared entry point for every enforcing service including ocl-ai-assistant.

core/capabilities/models.py
```diff
@@ -0,0 +1,88 @@
...
+
+class GroupCapability(models.Model):
+    """This group's limit for a capability. limit=0, or no row, = unlimited."""
+    class Meta:
+        db_table = 'group_capabilities'
+        unique_together = ('group', 'capability')
+
+    group = models.ForeignKey('auth.Group', on_delete=models.CASCADE, related_name='capabilities')
+    capability = models.ForeignKey(Capability, on_delete=models.CASCADE, related_name='group_capabilities')
+    limit = models.IntegerField()
```
paynejd: **B4.** `limit=0` = unlimited inverts the kill switch OpenConceptLab/ocl_issues#2782 specifies ("set the default role's caps to 0"), and leaves no way to express a zero allowance -- blocking requires deleting the row, which is indistinguishable from a misconfiguration.

Suggest `null=True` meaning unlimited and `0` meaning zero allowance, so the operator instruction in the ticket does what it says:

```suggestion
class GroupCapability(models.Model):
    """This group's limit for a capability. NULL = unlimited; 0 = no allowance (kill switch)."""
    class Meta:
        db_table = 'group_capabilities'
        unique_together = ('group', 'capability')

    group = models.ForeignKey('auth.Group', on_delete=models.CASCADE, related_name='capabilities')
    capability = models.ForeignKey(Capability, on_delete=models.CASCADE, related_name='group_capabilities')
    limit = models.IntegerField(null=True, blank=True)
```

`get_capability_limit`, `HasMapProjectCapacity`, the `rows_limit != 0` check in `map_projects/views.py` and `UserCapabilityOverrideSerializer(min_value=0)` all move with it.

core/map_projects/views.py
```diff
@@ -223,5 +277,34 @@ def get_object(self, queryset=None):  # pylint: disable=arguments-differ
...
+        completed_rows = serializer.validated_data.get('completed_rows')
+        completed_rows_delta = max((completed_rows or 0) - instance.completed_rows, 0)
+        try:
+            with transaction.atomic():
+                self.perform_update(serializer)
+                if completed_rows_delta:
+                    request.user.check_and_consume_capability(
+                        MAPPER_ROWS_PER_PROJECT_CAPABILITY_ID, units=completed_rows_delta,
+                        action='complete_automatch_rows', map_project=instance.map_project, run=instance
+                    )
```
paynejd: **S1.** `UsageCounter` is unique on `(user, capability)`, so consuming `mapper.rows_per_project` here accumulates across **all** of this user's projects -- but OpenConceptLab/ocl_issues#2762 scopes this cap to the project, and `AutomatchRunListView.post` enforces it the other way, against `map_project.rows_used`.

For a preview user with exactly one project the two agree by accident. For any grandfathered user they don't: OpenConceptLab/ocl_issues#2781 records 9 of 13 creators with more than one project, and they'd hit a single shared 25-row ceiling spanning all of them.

`MapProject.rows_used` already derives this per project with no counter -- I think this consume call should just go, leaving the declare-time check in `AutomatchRunListView.post` as the single enforcement point. If a completion-time meter is wanted too, it needs a project-scoped counter rather than the user ledger.

core/users/migrations/0036_seed_mapper_permissions_and_default_group.py
```diff
@@ -0,0 +1,69 @@
...
+    preview_group.permissions.set(
+        [permissions_by_codename[codename] for codename in PREVIEW_GROUP_PERMISSION_CODENAMES]
+    )
```
paynejd: **B1.** The group and its permissions are created, but no user is ever added to it in production code -- the only `groups.add(preview)` calls in this PR are in tests. Combined with `$match` now requiring `users.mapper_use`, this takes prod from 14 working Mapper users to 0.

OpenConceptLab/ocl_issues#2780 asks for the default role on user creation **and** a re-runnable backfill for the 1,459 existing accounts; OpenConceptLab/ocl_issues#2781 asks for the 14 `mapper-approved` users to be grandfathered into unlimited access before the gate opens. Both need to be in place before this merges.

Note B3 as well: a backfill that only writes Django group rows is undone at each user's next SSO login.

paynejd (changes requested):
*Supplement to my earlier review on this PR -- the blockers there stand and I'm not restating them. Everything below is additional, and the production figures are offered as evidence for the backfill blocker already raised.*

## One more blocker: SSO login erases `preview` membership

OpenConceptLab/ocl_issues#2780 flagged this as a hazard to verify before anything relies on it, and I think it's unaddressed. `OCLOIDCAuthenticationBackend.update_user` runs on **every** login and calls `set_groups(claims.get('groups', []))`:

```python
# core/common/backends.py:80
user.set_groups(claims.get('groups', []))

# core/users/models.py:310-315
def set_groups(self, groups, verify=True):
    if not verify or sorted(self.groups.values_list('name', flat=True)) != sorted(groups):
        self.groups.set(Group.objects.filter(name__in=groups))
```

`groups.set()` is a full replacement from the Keycloak claim. So membership in `preview` is durable only if Keycloak emits it, and the obvious manual remedy for the missing backfill -- adding users to `preview` in Django admin -- survives only until each user's next login, then silently vanishes along with their Mapper access.

This means the backfill can't be Django-only. It needs the Keycloak group to exist and be emitted in the claim, or `set_groups` needs to stop clobbering locally-granted groups. Since OCL supports standalone and Keycloak-integrated permanently (OpenConceptLab/ocl_issues#2780), both paths need to work before the gate opens. Neither the Keycloak configuration nor a `set_groups` change is in this PR.

## Production figures

Checked against the prod database, in support of the access blocker:

```
 total active users     1459
 mapper-waitlist         420
 mapper-approved          14     <- the only users who can currently use $match
 preview                absent
 auth_group pks         exactly 1-15, so pk 16 is free today
```

The `PREVIEW_GROUP_PK = 16` rationale does hold against prod right now. The risk is only that `get_or_create(pk=16, defaults={'name': 'preview'})` silently adopts and overwrites whatever group 16 is in an environment where that pk is taken -- keying on `name` removes the class of problem at no cost.

## Additional items

**`mapper.projects` usage drifts from what's actually enforced.** `HasMapProjectCapacity` compares against `map_projects_used` -- a live count, so deleting a project frees the slot, per the AC. But `GET /user/?includeCapabilities=true` reports `UsageCounter`, which is monotonic and never refunded. A user who deletes their one project sees `1/1 used` in the quota chip while creation actually succeeds. The Mapper's quota UI (OpenConceptLab/ocl_issues#2773) reads the second number and enforcement uses the first; they need to be the same number.

**The project cap is a TOCTOU.** `HasMapProjectCapacity.has_permission` reads the count outside any transaction and `create()` consumes after `save()`, so two concurrent creates both pass. Lower-volume than `$match`, but cheap to close.

**`AutomatchRunView` now accepts `PUT`.** The comment explaining why it was PATCH-only -- "the run-start snapshot is immutable, so a whole-object PUT has no meaning here" -- was replaced rather than answered. What needs `PUT`?

**`CapabilityExceeded` with `limit=None` renders as `"limit": null`.** An unconfigured user gets told they hit a quota. Worth a distinct `error_code` so the UI can say "not configured" rather than "limit reached" -- especially while the backfill question is open, since that is exactly the state affected users will be in.

**Coverage.** `MetadataToConceptsListView` carries `# pragma: no cover`, so the new quota path in `post()` is excluded, and OpenConceptLab/ocl_issues#2782 asks for tests in both standalone and Keycloak-replicated modes.

**PR metadata.** The title carries no ticket reference and the body lists four tickets under "Refs" with no closing keyword, so none will auto-close. Per our convention this wants the `OpenConceptLab/ocl_issues#N | ...` title form. Worth considering a split too -- OpenConceptLab/ocl_issues#2780 (permission layer) and OpenConceptLab/ocl_issues#2782 (enforcement) are separable, and OpenConceptLab/ocl_issues#2781 isn't implemented here at all.

core/concepts/views.py
```diff
@@ -1100,16 +1141,27 @@ def get_repo_params(is_semantic, target_repo_params, target_repo_url):
...
+                )
+            except CapabilityExceeded as ex:
+                return Response(
+                    {
+                        'detail': 'Match operation limit reached.',
+                        'error_code': CAPABILITY_EXCEEDED_ERROR_CODE[MAPPER_MATCH_OPERATIONS_CAPABILITY],
+                        'limit': ex.limit, 'used': ex.used,
+                    },
+                    status=status.HTTP_403_FORBIDDEN
+                )
```
paynejd, line 1173: **An unentitled user is told they've exhausted an allowance they never had — and on deploy day that's everyone.**

`get_capability_limit()` returns `None` when there's no per-user override *and* no `GroupCapability` row (`core/users/models.py:268-270`), and `check_and_consume_capability` correctly treats that as **blocked** — its docstring is explicit: *"A capability with no override and no group row for this user at all ... is BLOCKED, not unlimited."*

But the `CapabilityExceeded` it raises carries `limit=None`, and this handler flattens two very different states into one message:

```python
'detail': 'Match operation limit reached.',
'error_code': CAPABILITY_EXCEEDED_ERROR_CODE[MAPPER_MATCH_OPERATIONS_CAPABILITY],
'limit': ex.limit, 'used': ex.used,     # -> null, 0
```

So a user who is simply **not entitled** receives `mapper_match_operations_limit_reached, limit: null, used: 0`, which oclmap renders through `PreviewLimitDialog` as *"Your free preview's match-operation limit has been reached"* (`preview_limit_body_match_operations_no_count`). Told they spent an allowance that was never provisioned.

This is not an edge case at launch. Because there's no #2781 backfill, `None` is the **default** state for every non-superuser on deploy — so the first thing every existing Mapper user sees is a quota-exhausted message.

I raised exactly this as a blocker on ocl-ai-assistant#195 (`core/common/permissions.py`, where I suggested an `ai_assistant_not_available` code). That fix is incomplete while this side behaves the same way, and it would leave the two services using different vocabulary for the same condition.

Suggest distinguishing them at the source — give `CapabilityExceeded` an `entitled` flag (or check `ex.limit is None` at each handler) and emit a distinct code:

```python
            except CapabilityExceeded as ex:
                if ex.limit is None:
                    # No override and no group row: an entitlement problem, not an
                    # exhausted quota. Don't quote cap numbers that don't exist.
                    return Response(
                        {
                            'detail': 'The Mapper is not available for this account.',
                            'error_code': 'mapper_not_entitled',
                        },
                        status=status.HTTP_403_FORBIDDEN
                    )
                return Response(
                    {
                        'detail': 'Match operation limit reached.',
                        'error_code': CAPABILITY_EXCEEDED_ERROR_CODE[MAPPER_MATCH_OPERATIONS_CAPABILITY],
                        'limit': ex.limit, 'used': ex.used,
                    },
                    status=status.HTTP_403_FORBIDDEN
                )
```

Same treatment needed at `core/map_projects/views.py:230` and `:301`, `core/capabilities/exceptions.py:32`, and `core/capabilities/views.py:59`. Whatever code you settle on needs a `CAP_COPY_KEY` entry in oclmap's `PreviewLimitDialog` and a matching branch in ai-assistant — I've filed both.

core/common/permissions.py
```diff
@@ -97,10 +102,112 @@ def has_object_permission(self, request, view, obj):
...
+
+
+class CanUseMapper(HasMapperCapability):
+    capability_name = MAPPER_USE_PERMISSION
+
+
+class CanUseMapperAIAssistant(HasMapperCapability):
+    capability_name = MAPPER_AI_ASSISTANT_PERMISSION
+    error_code = 'mapper_ai_assistant_denied'
+    denied_message = 'The AI Assistant is not available for your account.'
```
paynejd: **`CanUseMapperAIAssistant` is defined but wired to no view, and ai-assistant answers the same denial with an untyped 403.**

No view in this PR references it:

```
$ git grep -n 'CanUseMapper\|CanUseCustomMapperAlgorithms\|CanCreateOrgMapProjects' -- core | grep -v common/permissions.py
core/concepts/views.py:929:     permission_classes = (IsAuthenticated, CanUseMapper)
core/concepts/views.py:1179:    permission_classes = (IsAuthenticated, CanUseMapper)
core/map_projects/views.py:35:  ... CanCreateOrgMapProjects, HasMapProjectCapacity, CanUseCustomMapperAlgorithms,
core/map_projects/views.py:39:  ... CanCreateOrgMapProjects, CanUseCustomMapperAlgorithms)
core/map_projects/views.py:44:  ... CanCreateOrgMapProjects)
```

`CanUseMapperAIAssistant` is absent — which makes sense, since the endpoint it would guard lives in ocl-ai-assistant. But over there the same permission is enforced as:

```python
# ocl-ai-assistant core/common/views.py:75-77  (PR #195)
if not is_authorized:
    raise Http403()          # no error_code, no detail
return True
```

So the carefully structured `error_code` contract stops at the service boundary. A user without `users.mapper_ai_assistant` gets a bare 403, oclmap finds no `error_code`, and it falls through to `t('unknown_error')` — while a perfectly good `mapper_ai_assistant_denied` sits unused here.

Pick one and make it real:

- **Preferred:** keep this class as the canonical definition and have ai-assistant's `check_permission` raise the same structured payload (`{'detail': ..., 'error_code': 'mapper_ai_assistant_denied'}`), so one vocabulary spans both services. I've noted the ai-assistant half on #195.
- Or drop `CanUseMapperAIAssistant` if nothing in oclapi2 will ever gate on it, and let ai-assistant own that code — but then it still has to *emit* one.

Either way oclmap needs a `CAP_COPY_KEY` entry for it; filed on #60. Same question applies to `mapper_custom_algorithms_denied`, which **is** reachable (map-project create/update) and which oclmap also doesn't map.

paynejd (commented):
## Cross-PR alignment pass — #900 / oclmap#60 / ocl-ai-assistant#195

I've now reviewed all three `mapper-preview` PRs. Reading them side by side surfaced two things that belong on **this** PR and that I missed by reviewing it in isolation, plus a sequencing note that was only on one of the three. Nothing here changes my earlier blockers.

**Deploy sequencing — all three must land together, in this order.** This was stated on ocl-ai-assistant#195 and on neither of the other two, so repeating it in one place on each PR:

1. **oclapi2#900** + the ocl_issues#2781 backfill — defines the capabilities, permissions, error codes and the `/capabilities/consume/` endpoint. Nothing else works without it, and without the backfill it takes Mapper access away from the 14 existing `mapper-approved` users.
2. **ocl-ai-assistant#195** — sends `includeCapabilities=true` and reads `permissions`/`capabilities`. Against an oclapi2 that doesn't serve them, both `check_permission` and the capacity pre-check fail closed: total ai-assistant outage.
3. **oclmap#60** — reads the meters and error codes both of the above produce.

Merging any of these independently produces a broken environment, not a partially-working one.

### Two items I raised on the wrong PR

**1. `limit is None` reported as "you've used your allowance" — I called this a blocker on ocl-ai-assistant#195, but the root cause is here.** Detail inline on `core/concepts/views.py`. It's the same defect in four places in this PR, and the #195 fix can't be correct on its own if this side keeps emitting `*_limit_reached` with `limit: null`.

**2. `mapper_ai_assistant_denied` is defined here and used by no view, while ocl-ai-assistant raises a bare `Http403()` for that exact denial.** Detail inline on `core/common/permissions.py`. Net effect today: an unentitled user gets an untyped 403 that oclmap renders as a generic error.

### Three of my original items are cross-repo decisions, not backend-only

Flagging so they're not fixed in one repo and left in the others:

- **Metering unit.** `units=len(rows)` here vs row-algorithm pairs in #2762/#2782. oclmap#60 implements the *spec* (it multiplies by `algorithmCount` in both the estimate and the pre-truncation), so the two are actively inconsistent today — the UI truncates runs this server would allow. ai-assistant's `units=1` per `$invoke` is unaffected either way.
- **`0 = unlimited`.** If the sentinel changes to resolve the inverted kill switch, it changes in three places: `get_capability_limit` here, `if limit and used >= limit` in ai-assistant's `HasAIAssistantCapacity`, and `unlimited = limit === 0` in oclmap's `toMeter`.
- **`None = blocked`.** Same three repos. oclmap currently gets this *backwards* in one helper (`getPreviewEligibleRowIndexes` treats a null limit as unlimited) — filed on #60.

Attribution (`map_project`/`run` dropped in `CapabilityConsumeView`) is already cross-linked with #195 in both directions; no change needed there.

core/users/migrations/0036_seed_mapper_permissions_and_default_group.py
snyaggarwal: Preview group is now handled via fixtures, this migration is not needed anymore

core/concepts/views.py
snyaggarwal: We now are refunding after usage.

this is true $match attribution -- if 25 rows are run against 4 algo then its 100 operations and thats what its counting. This is the simplistic view. We will have to adjust the starting value.

core/users/constants.py
snyaggarwal: holding on until we discuss 2781

core/fixtures/capabilities.yaml
snyaggarwal: added a `seed_group_capabilities` which will run on startup and only do something if it doesnt exist.

core/common/permissions.py
snyaggarwal: moved to ocl-ai-assistant

snyaggarwal:
@paynejd review applied

core/capabilities/views.py
```diff
@@ -0,0 +1,223 @@
...
+
+        run = None
+        run_id = request.data.get('run', None)
+        if isinstance(run_id, (str, int)) and str(run_id).isdigit():
+            run_qs = AutomatchRun.objects.filter(id=int(run_id))
+            run_qs = run_qs.filter(map_project=map_project) if map_project else run_qs.filter(
+                map_project__created_by=request.user)
+            run = run_qs.first()
+
+        return map_project, run
```
paynejd: **Blocking: a crafted id makes AI calls free.** `'²'.isdigit()` is True but `int('²')` raises, so a `map_project` or `run` of `"²"` makes this view 500 before anything is consumed. ai-assistant takes those ids from `X-OCL-Event-Metadata`, which any user can set, consumes only after the LLM call, and treats the 500 as a network error, so the caller gets the result unmetered. A verifier reproduced it end to end: 8 × 200, `used` frozen.

```suggestion
    @staticmethod
    def parse_id(value):
        # Plain ASCII digits within bigint range only: '²'.isdigit() is True but int('²') raises (a 500).
        if isinstance(value, int) and not isinstance(value, bool):
            value = str(value)
        if isinstance(value, str) and value.isascii() and value.isdigit() and len(value) <= 18:
            return int(value)
        return None

    def get_owned_map_project_and_run(self):
        from core.map_projects.models import AutomatchRun, MapProject

        request = self.request
        map_project = None
        map_project_id = self.parse_id(request.data.get('map_project', None))
        if map_project_id is not None:
            # .only('id'): these rows only feed the UsageEvent FKs
            map_project = MapProject.objects.filter(id=map_project_id, created_by=request.user).only('id').first()

        run = None
        run_id = self.parse_id(request.data.get('run', None))
        if run_id is not None:
            run_qs = AutomatchRun.objects.filter(id=run_id)
            run_qs = run_qs.filter(map_project=map_project) if map_project else run_qs.filter(
                map_project__created_by=request.user)
            run = run_qs.only('id').first()

        return map_project, run
```

Anything that isn't plain ASCII digits within bigint range is ignored, so the event is just unattributed. `.only('id')` because these rows only set FKs.

core/concepts/views.py
```diff
@@ -880,12 +885,52 @@ def post(request):
...
+
+    map_project = None
+    map_project_id = metadata.get('map_project_id')
+    if isinstance(map_project_id, (str, int)) and str(map_project_id).isdigit():
+        from core.map_projects.models import MapProject
+        # Scoped to request.user: the id is client-supplied (this header), so without
+        # this a caller could attribute their UsageEvent to a project they don't own,
+        # corrupting per-project attribution even though usage still charges the right
+        # user's quota.
+        map_project = MapProject.objects.filter(id=int(map_project_id), created_by=request.user).first()
```
paynejd: **Blocking: the same parse, plus a full-row load on every `$match` batch.** A `map_project_id` of `"²"` 500s the `$match` here. That's not a bypass, since nothing has been consumed yet. And `.first()` loads the whole `MapProject` just to set the `UsageEvent` FK. On prod the largest project stores 68 MB across `candidates`, `matches`, `analysis` and `logs`, 46 of 99 store over 1 MB, and this runs for every batch of every run.

```suggestion
    map_project_id = metadata.get('map_project_id')
    if isinstance(map_project_id, int) and not isinstance(map_project_id, bool):
        map_project_id = str(map_project_id)
    # Plain ASCII digits within bigint range only: '²'.isdigit() is True but int('²') raises (a 500).
    if (isinstance(map_project_id, str) and map_project_id.isascii() and map_project_id.isdigit()
            and len(map_project_id) <= 18):
        from core.map_projects.models import MapProject
        # Scoped to request.user: the id is client-supplied (this header), so without
        # this a caller could attribute their UsageEvent to a project they don't own,
        # corrupting per-project attribution even though usage still charges the right
        # user's quota. .only('id'): the row only sets the UsageEvent FK, so don't load
        # candidates/matches/analysis/logs on every $match batch.
        map_project = MapProject.objects.filter(id=int(map_project_id), created_by=request.user).only('id').first()
```

Same rule as the consume view; sharing its helper works too.

core/map_projects/views.py
```diff
@@ -178,6 +236,39 @@ def get(self, request, *args, **kwargs):
...
+                    return Response(
+                        {
+                            'detail': 'You do not have a configured row allowance for this project.'
+                            if not_entitled else 'Preview row limit for this project reached.',
+                            'error_code': CAPABILITY_NOT_ENTITLED_ERROR_CODE[MAPPER_ROWS_PER_PROJECT_CAPABILITY]
+                            if not_entitled else CAPABILITY_EXCEEDED_ERROR_CODE[MAPPER_ROWS_PER_PROJECT_CAPABILITY],
+                            'limit': rows_limit, 'used': rows_used,
+                        },
+                        status=status.HTTP_403_FORBIDDEN
+                    )
```
paynejd: **Blocking: `mapper.rows_per_project` is metered as a quota, but it's a cap.** A cap is how many of something may exist: database state, freed on delete. A quota is how many operations you may perform in the period. `rows_used + intended_rows` sums the rows of every Auto Match run on the project (retries aside), so a preview user gets one bulk pass per project, and every later run is refused while match quota remains.

```suggestion
        intended_rows = serializer.validated_data['intended_rows']
        is_retry = bool(serializer.validated_data.get('parent_run'))

        # mapper.match_operations is metered per row-algorithm pair at $match
        # time (MetadataToConceptsListView.post()), not here - a run's actual
        # $match calls (one per algorithm per row-batch, fired as the run
        # executes) already account for every unit this run will use. Consuming
        # it again at run creation double-counted the whole run.
        # mapper.rows_per_project is a cap on project size, not a quota: each run
        # is checked on its own, so re-running Auto Match costs match quota only.
        # 0 = unlimited, -1 = blocked, None = not entitled (blocked).
        if not is_retry:
            rows_limit = request.user.get_capability_limit(MAPPER_ROWS_PER_PROJECT_CAPABILITY_ID)
            if rows_limit != 0 and (rows_limit is None or intended_rows > rows_limit):
                not_entitled = rows_limit is None or rows_limit < 0  # -1 = blocked: "not available", no -1 count
                return Response(
                    {
                        'detail': 'You do not have a configured row allowance for this project.'
                        if not_entitled else 'This run is larger than the preview row limit for a project.',
                        'error_code': CAPABILITY_NOT_ENTITLED_ERROR_CODE[MAPPER_ROWS_PER_PROJECT_CAPABILITY]
                        if not_entitled else CAPABILITY_EXCEEDED_ERROR_CODE[MAPPER_ROWS_PER_PROJECT_CAPABILITY],
                        'limit': None if not_entitled else rows_limit, 'requested': intended_rows,
                    },
                    status=status.HTTP_403_FORBIDDEN
                )
```

Each run is checked on its own, so re-running costs match quota only. The complete cap enforces project size as state at import: count the uploaded file's rows on create/update. That can follow; until then, compute is bounded by the match quota. I dropped the `map_project` local, since nothing else reads it. This leaves `MapProject.rows_used` unused, and it flips `test_post_rows_per_project_accumulates_across_top_level_runs` (that's the quota behavior itself) and the `used` assertion in `test_post_rows_per_project_limit_reached`.

core/map_projects/views.py
```diff
@@ -223,5 +314,33 @@ def get_object(self, queryset=None):  # pylint: disable=arguments-differ
...
+                # its already-approved rows needs no second gate here. Using
+                # check_and_consume_capability here (a per-user, never-reset counter)
+                # previously meant a user with multiple projects could get permanently
+                # locked out of progress on ALL of them once their lifetime total crossed
+                # the limit, even on a brand new project with zero rows. This only logs
+                # the event for attribution/reporting.
+                request.user.log_capability_usage(
+                    MAPPER_ROWS_PER_PROJECT_CAPABILITY_ID, units=completed_rows_delta,
+                    action='complete_automatch_rows', map_project=instance.map_project, run=instance
+                )
```
paynejd: **A cap has no usage.** This bumps a per-user, lifetime `UsageCounter` for `mapper.rows_per_project`, and `/user/?includeCapabilities=true` reports that counter as `used`. oclmap subtracts it and truncates every Auto Match to what's left, so once a user has completed 25 rows anywhere, Auto Match processes zero rows on every project.

```suggestion
        """Persist lifecycle progress. mapper.rows_per_project is a cap, so completed rows aren't metered."""
        partial = kwargs.pop('partial', False)
        instance = self.get_object()
        serializer = self.get_serializer(instance, data=request.data, partial=partial)
        serializer.is_valid(raise_exception=True)
        self.perform_update(serializer)
```

The `AutomatchRun` already records completed rows per run, and `log_capability_usage` has no other caller. Likewise, [`get_capability_usage`](https://github.com/OpenConceptLab/oclapi2/blob/3a15b6dbbbb15f0ffec7a6ac12159995ecad8a3d/core/users/models.py#L280-L288) shouldn't report a per-user `used` for a per-project cap: return `None` for it so clients show the cap itself. `test_patch_progress` and `test_patch_progress_not_reblocked_by_a_later_lowered_override` assert the counter and the event, so they change with this.

core/capabilities/management/commands/seed_group_capabilities.py
```diff
@@ -0,0 +1,96 @@
...
+    if missing:
+        warn(f'{group_name}: unknown permissions skipped: {sorted(missing)}')
+    group.permissions.add(*permissions)
+
+    for capability_name, limit in (group_config.get('capabilities') or {}).items():
+        capability = Capability.objects.filter(name=capability_name).first()
+        if not capability:
+            warn(f'{group_name}: unknown capability skipped: {capability_name}')
+            continue
+        GroupCapability.objects.get_or_create(group=group, capability=capability, defaults={'limit': limit})
```
paynejd: **Blocking: the seed is add-only, so the config can't lower a limit or revoke a permission.** `get_or_create` keeps whatever limit exists, and `permissions.add` never removes. It also undoes today's documented kill switch: delete the `GroupCapability` row, and the next restart recreates it. The groups config (OpenConceptLab/oclinfrastructure#7) is the source of truth for every group it declares, so make this a sync:

```suggestion
    group, _ = Group.objects.get_or_create(name=group_name)
    # sync: the group's users.* permissions become exactly the listed ones; other apps' permissions are left alone
    codenames = group_config.get('permissions') or []
    permissions = list(Permission.objects.filter(codename__in=codenames, content_type__app_label='users'))
    missing = set(codenames) - {permission.codename for permission in permissions}
    if missing:
        warn(f'{group_name}: unknown permissions skipped: {sorted(missing)}')
    group.permissions.set([*group.permissions.exclude(content_type__app_label='users'), *permissions])

    # sync: declared limits are upserted and undeclared ones deleted (no row = not entitled)
    declared_ids = []
    for capability_name, limit in (group_config.get('capabilities') or {}).items():
        capability = Capability.objects.filter(name=capability_name).first()
        if not capability:
            warn(f'{group_name}: unknown capability skipped: {capability_name}')
            continue
        GroupCapability.objects.update_or_create(group=group, capability=capability, defaults={'limit': limit})
        declared_ids.append(capability.id)
    GroupCapability.objects.filter(group=group).exclude(capability_id__in=declared_ids).delete()
```

Groups the config doesn't declare stay untouched, and so do other apps' permissions. The kill switch becomes: set the limit to `-1` in groups.yaml and restart.

core/capabilities/management/commands/seed_group_capabilities.py
```diff
@@ -0,0 +1,96 @@
...
+import os
+
+import yaml
+from django.core.management import BaseCommand, CommandError
+from django.db import transaction
+
+GROUPS_CONFIG_ENV = 'OCL_GROUPS_CONFIG'
+
+# Plan groups are deployment config: {"groups": {"<name>": {"permissions": [...], "capabilities": {...}}}}
+# Seed-only - creates missing groups/limits and adds missing permissions; never overwrites or removes.
```
paynejd: Header to match:

```suggestion
# Plan groups are deployment config: {"groups": {"<name>": {"permissions": [...], "capabilities": {...}}}}
# Sync - the config is the source of truth for every group it declares: users.* permissions are set exactly,
# declared limits are upserted and undeclared ones deleted. Groups it doesn't declare are left alone.
# Limits: 0 = unlimited, -1 = blocked (the kill switch), N > 0 = the limit; no row = not entitled (blocked).
```

core/capabilities/management/commands/seed_group_capabilities.py
```diff
@@ -0,0 +1,96 @@
...
+                not isinstance(capabilities, dict) or
+                not all(isinstance(limit, int) and not isinstance(limit, bool) for limit in capabilities.values())
+        ):
+            raise CommandError(f'{group_name}: capabilities must map names to integer limits')
+    return config['groups']
+
+
+class Command(BaseCommand):
+    help = f'seed groups, their permissions and capability limits from {GROUPS_CONFIG_ENV} or --file, ' \
+           'without overwriting any that already exist'
```
paynejd: And the help text:

```suggestion
    help = f'sync groups, their users.* permissions and capability limits from {GROUPS_CONFIG_ENV} or --file; ' \
           'the config is the source of truth for every group it declares'
```

core/capabilities/tests/tests.py
```diff
@@ -0,0 +1,355 @@
...
+                'pro': {
+                    'permissions': ['mapper_use', 'mapper_ai_assistant'],
+                    'capabilities': {'mapper.projects': 99, 'ai_assistant.calls': 50},
+                }
+            }
+        })
+
+        self.assertEqual(self.codenames('pro'), {'mapper_use', 'mapper_custom_algorithms', 'mapper_ai_assistant'})
+        self.assertEqual(
+            self.limits('pro'), {'mapper.projects': 20, 'mapper.match_operations': 0, 'ai_assistant.calls': 50})
```
paynejd: With the sync, this test flips. Using `-1` here also exercises the kill switch:

```suggestion
    def test_reseed_syncs_permissions_and_limits(self):
        self.seed(self.CONFIG)
        self.seed({
            'groups': {
                'pro': {
                    'permissions': ['mapper_use', 'mapper_ai_assistant'],
                    'capabilities': {'mapper.projects': -1, 'ai_assistant.calls': 50},
                }
            }
        })

        self.assertEqual(self.codenames('pro'), {'mapper_use', 'mapper_ai_assistant'})
        self.assertEqual(self.limits('pro'), {'mapper.projects': -1, 'ai_assistant.calls': 50})
```

core/capabilities/models.py
```diff
@@ -0,0 +1,88 @@
...
+
+    name = models.CharField(max_length=100, unique=True)
+    description = models.TextField(blank=True, default='')
+
+    def __str__(self):
+        return str(self.name or self.__class__.__name__)
+
+
+class GroupCapability(models.Model):
+    """This group's limit for a capability. limit=0, or no row, = unlimited."""
```
paynejd: **This docstring contradicts the code.** No row doesn't mean unlimited: `get_capability_limit` returns `None`, and consumption blocks. It's also where people will look for the kill switch.

```suggestion
    """
    This group's limit for a capability: 0 = unlimited, -1 = blocked (the kill switch), N > 0 = the limit.
    No row = this group grants nothing; a user with no row anywhere is not entitled (blocked).
    """
```

[`core/users/models.py:300`](https://github.com/OpenConceptLab/oclapi2/blob/3a15b6dbbbb15f0ffec7a6ac12159995ecad8a3d/core/users/models.py#L300) and [`core/common/permissions.py:206`](https://github.com/OpenConceptLab/oclapi2/blob/3a15b6dbbbb15f0ffec7a6ac12159995ecad8a3d/core/common/permissions.py#L206) still name row deletion as the kill switch. A deleted row comes back on restart, so they want this wording too. Separately, the 403s report `-1` as `*_limit_reached` with `limit: -1`, which oclmap renders as a count. Report it as not available (`limit: null`), as OpenConceptLab/ocl-ai-assistant#195 is being asked to.

core/capabilities/models.py
```diff
@@ -0,0 +1,88 @@
...
+        db_table = 'group_capabilities'
+        unique_together = ('group', 'capability')
+
+    group = models.ForeignKey('auth.Group', on_delete=models.CASCADE, related_name='capabilities')
+    capability = models.ForeignKey(Capability, on_delete=models.CASCADE, related_name='group_capabilities')
+    limit = models.IntegerField()
+
+
+class UserCapabilityOverride(models.Model):
+    """Per-user override, wins over every GroupCapability the user's groups carry. limit=0 = unlimited."""
```
paynejd: Same sentinels for overrides; `-1` is how staff block a single user:

```suggestion
    """
    Per-user override; wins over every GroupCapability the user's groups carry.
    0 = unlimited, -1 = blocked (this user only), N > 0 = the limit.
    """
```

core/capabilities/serializers.py
```diff
@@ -0,0 +1,18 @@
...
+    def get_limit(self, obj):
+        return self.context['user'].get_capability_limit(obj.id)
+
+    def get_used(self, obj):
+        return self.context['user'].get_capability_usage(obj.id)
+
+
+class UserCapabilityOverrideSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    capability = CapabilitySerializer(read_only=True)
+    limit = serializers.IntegerField(min_value=0)
```
paynejd: `min_value=0` stops staff from blocking a single user with an override. Allow `-1`:

```suggestion
    limit = serializers.IntegerField(min_value=-1)  # 0 = unlimited, -1 = blocked, N > 0 = the limit
```

`test_put_rejects_negative_limit` then wants `-2`.

core/map_projects/views.py
```diff
@@ -45,17 +77,43 @@ def get(self, request, *args, **kwargs):
     def create(self, request, **kwargs):  # pylint: disable=unused-argument
         if not self.parent_resource:
             return Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)
-        permission = HasOwnership()
-        if not permission.has_object_permission(request, self, self.parent_resource):
-            return Response(status=status.HTTP_403_FORBIDDEN)
+
         serializer = self.get_serializer(data=MapProject.format_request_data(request.data, self.parent_resource))
```
paynejd: **Blocking: ownership comes from the request body.** `format_request_data` keeps a client-sent `organization_id` next to the URL parent's id, and `MapProject.parent` prefers the org. So `POST /users/<me>/map-projects/` with `organization_id=<any org>` creates the project in that org, past the new org gate and org membership. It predates this PR (OpenConceptLab/ocl_issues#2804), but this PR's gate relies on it being closed.

```suggestion
        data = MapProject.format_request_data(request.data, self.parent_resource)
        # Ownership comes only from the URL parent: a body organization_id posted to a user route would re-parent
        # the project past HasMapProjectParentOwnership and CanCreateOrgMapProjects (OpenConceptLab/ocl_issues#2804).
        parent_key = self.parent_resource.resource_type.lower() + '_id'
        for key in ('organization_id', 'user_id'):
            if key != parent_key:
                data.pop(key, None)
        serializer = self.get_serializer(data=data)
```

`format_request_data` returns a new dict, so this never touches `request.data` or the uploaded file. Updates are already safe, since `prepare_object` only sets ownership on new instances, and oclmap posts org projects to the org URL, so no real client sends these fields. It still wants regression tests on both routes, per the ticket's ACs.

core/users/serializers.py
```diff
@@ -253,14 +270,7 @@ def update(self, instance, validated_data):
...
             if isinstance(auth_groups, list):
-                if len(auth_groups) == 0:
-                    instance.groups.set([])
-                else:
-                    if instance.is_valid_auth_group(*auth_groups):
-                        instance.groups.set(Group.objects.filter(name__in=auth_groups))
-                    else:
-                        self._errors.update({'auth_groups': [INVALID_AUTH_GROUP_NAME]})
-                        return instance
+                instance.groups.set(Group.objects.filter(name__in=auth_groups))
```
paynejd: **Blocking: in standalone mode, any user can put themselves in any group.** a5fa9c5b dropped the `AUTH_GROUPS` allow-list, so `PUT /user/?includeAuthGroups=true` with `{"auth_groups": ["core_user"]}` grants the caller whatever that group carries, including the unlimited plan groups. Restoring the allow-list wouldn't fix it: it already listed `core_user` and `staff_user`, which groups.yaml makes unlimited.

```suggestion
        # Standalone only (under SSO, Keycloak owns membership via set_groups), and admin-only: a group carries
        # permissions and capability limits, so letting users pick their own groups lets them pick their own limits.
        if not AuthService.is_sso_enabled() and request_user.is_staff:
            auth_groups = validated_data.get('auth_groups', None)
            if isinstance(auth_groups, list):
                instance.groups.set(Group.objects.filter(name__in=auth_groups))
```

`auth_groups` from non-staff is ignored, as it already is under SSO. It wants a test that a non-staff self-PUT leaves groups alone.

core/capabilities/views.py
```diff
@@ -0,0 +1,223 @@
...
+        if error_response:
+            return error_response
+        map_project, run = self.get_owned_map_project_and_run()
+
+        try:
+            request.user.check_and_consume_capability(
+                capability_id, units=units,
+                action=request.data.get('action', ''), algorithm=request.data.get('algorithm'),
+                map_project=map_project, run=run,
+            )
```
paynejd: **Blocking: consume needs idempotency.** To rule out overspend, ai-assistant will consume before the LLM call and fail closed, and oclmap will send a stable per-(run, row) key (the matching changes are on OpenConceptLab/ocl-ai-assistant#195 and OpenConceptLab/oclmap#60). Without dedup here, every retry of the same row charges again. This needs a migration, so no suggestion:

- Accept an optional `idempotency_key` in the body.
- Add a nullable `UsageEvent.idempotency_key`, with a partial unique constraint on (user, capability, idempotency_key) where the key is set.
- In `check_and_consume_capability`, look the key up under the counter's `select_for_update` and before the limit check. If a matching `UsageEvent` exists, return 200 with that event and `already_consumed: true`, and charge nothing. Checking before the limit means a retry after the last unit still gets its 200.

ai-assistant then returns its stored result instead of calling the LLM again.

paynejd (changes requested):
Thanks, Sunny. The round-1 fixes all landed: the re-create 500, negative `units`, the pk-16 group, org projects going read-only, staff 403s, the projects meter, the create race, the not-entitled codes, the `/user/map-projects/` 403 and the attribution ownership checks.

**Blocking**

1. **Attribution ids** (`core/capabilities/views.py:53`). `'²'.isdigit()` is True but `int('²')` raises, so `/capabilities/consume/` returns a 500. ai-assistant treats that as a network error and returns the LLM result unmetered. Any user can set `X-OCL-Event-Metadata`; a verifier reproduced it end to end (8 × 200, `used` frozen). The `$match` parse has the same bug.
2. **Full-row load** (`core/concepts/views.py:916`). `$match` attribution loads the whole `MapProject` (up to 68 MB on prod) on every batch just to set an FK. Use `.only('id')`.
3. **Row cap metered as a quota** (`core/map_projects/views.py:240`, `:318`). `mapper.rows_per_project` is a running sum of Auto Match rows per project: one bulk pass, then re-runs are refused while match quota remains. `/user/` also reports a per-user lifetime counter as `used`. It's a cap.
4. **Add-only seed** (`seed_group_capabilities.py:29`). It can never lower a limit or revoke a permission, and a deleted row (today's documented kill switch) comes back on restart. Make it sync and document `-1` = blocked (`core/capabilities/models.py:17`).
5. **Org bypass** (`core/map_projects/views.py:81`). A body `organization_id` bypasses the org gate and org membership (OpenConceptLab/ocl_issues#2804). It predates this PR, but this PR's gate relies on it being closed.
6. **Standalone self-assignment** (`core/users/serializers.py:270`). Since a5fa9c5b, any user can `PUT /user/?includeAuthGroups=true` into any group, including unlimited ones.
7. **Consume idempotency** (`core/capabilities/views.py:91`), so ai-assistant can consume before invoking without double-charging retries.

**Decisions**

- **Caps vs quotas.** A cap is how many of something may exist (database state, freed on delete). A quota is how many operations you may perform in the period (preview → November). Projects, match operations and AI calls comply; rows per project doesn't yet.
- **Keycloak rule.** With Keycloak, groups, roles and permissions live in Keycloak and overwrite Django, so `set_groups` is correct and I withdraw my round-1 concern. The groups config is the source of truth and the seed syncs; no Django-admin tuning. Standalone membership is admin-only.
- **No overspend.** The AI path moves to consume-first and fail-closed; this PR provides the idempotent consume.

**Deploy order** (please add it to the PR description): OpenConceptLab/oclinfrastructure#7 → OpenConceptLab/oclapi2#900 → Keycloak groups & membership → OpenConceptLab/ocl-ai-assistant#195 → OpenConceptLab/oclmap#60. Please also retitle to `OpenConceptLab/ocl_issues#2762 | ...` with full-reference `Refs`/`closes` lines; the bare "Refs" list closes nothing.

**Follow-ups (non-blocking)**

- `?includeCapabilities=true` shows any user's limits and usage to any logged-in user, and adds ~15 queries per user on list views.
- `UsageEvent` isn't reversed when `$match` refunds.
- `AnalyticsMiddleware` records `/capabilities/consume/` as end-user traffic, double-counting AI activity in `api_transactions` (the pricing data).
- No tests for `/capabilities/consume|refund/` or Keycloak mode.
- `# pragma: no cover` still hides the `$match` quota path.
- The override serializer can't set `-1` (one-click fix inline).

